PR #1, opened 2019-03-20
vmm-sys-util

liujing2:
This includes the utilities that directly use libc from crosvm and firecracker.


src/eventfd.rs
```diff
@@ -0,0 +1,140 @@
...
+/// A safe wrapper around a Linux eventfd (man 2 eventfd).
+///
+/// An eventfd is useful because it is sendable across processes and can be used for signaling in
+/// and out of the KVM API. They can also be polled like any other file descriptor.
+pub struct EventFd {
+    eventfd: File,
+}
+
+impl EventFd {
+    /// Creates a new blocking EventFd with an initial value of 0.
```
andreeaflorescu: I don't think this comment is accurate. We are creating the eventfd with EFD_NONBLOCK.
I noticed that in CrosVM they are opening it without the flag so maybe it would make sense to pass the additional flags as parameters.

src/errno.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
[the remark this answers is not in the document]
zachreizner: The nice part about this error type is that it is fully compatible with os error code thrown off by system calls. If one were to put an os error code into an `io::Error`, it might be hard to get the os error back out in upper parts of the call stack. Using `io::Error::raw_os_error` returns an `Option` that either requires `unwrap` or an extra `match` just to extract the code. Using `io::Error::kind` is tricky because it has variants which are not possible to get from a syscall (`UnexpectedEof`), or are mismatched by the std implementation that converts error numbers to `ErrorKind` variants. Additionally, some error codes have no correspondence with an `ErrorKind` variant other than `Other`.

If human error messaging is all that is needed to make this module acceptable, I suggest using what we have in crosvm today for `Error`:

```
impl Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        io::Error::from_raw_os_error(self.0).fmt(f)
    }
}
```

andreeaflorescu:
@liujing2 please also add the Travis configuration file. Here is the one from the vm-memory crate: https://github.com/rust-vmm/vm-memory/blob/master/.travis.yml

src/errno.rs
```diff
@@ -0,0 +1,86 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO: This module is not used now. Avoid the warning.
```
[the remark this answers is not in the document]
liujing2: Thank you for the quick reviewing! I'll fix the comments in codes.

README.md
```diff
@@ -1 +1,2 @@
-# vmm-sys-util
\ No newline at end of file
+# vmm-sys-util
+This crate is a collection of modules that essentially provide helpers and utilities.
```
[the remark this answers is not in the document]
liujing2: Yes. Will do it.

src/eventfd.rs
```diff
@@ -0,0 +1,140 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//TODO: Firecracker uses io::Error instead of sys_util::Error for Result<T, E>. Need discussion here.
+
+#![allow(unused)]
```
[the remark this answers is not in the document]
liujing2: Right! Will remove it

src/eventfd.rs
liujing2: Yep. Let me change the flag to be a parameter for new() and test the new(EFD_NONBLOCK) in #[test]. Do you think we need double test the new(0) for each #[test]?

andreeaflorescu: Hmm, I'll look at the code. It is always nice to test all the usecases, now it depends whether it adds any value.

src/timerfd.rs
```diff
@@ -0,0 +1,163 @@
...
+use std::time::Duration;
+
+use libc::{self, timerfd_create, timerfd_gettime, timerfd_settime, CLOCK_MONOTONIC, TFD_CLOEXEC};
+
+use crate::errno::{errno_result, Result};
+
+/// A safe wrapper around a Linux timerfd (man 2 timerfd_create).
+pub struct TimerFd(File);
+
+impl TimerFd {
```
andreeaflorescu: I haven't check the other docs, but I think it is a good idea to have the documentation split in summary and description.

For new() it would look something like this:
```rust
/// Create a new `TimerFd`.
///
/// The timer is initially disarmed and must be armed by calling `reset()`.
```

Then we could also add links to the structures and methods, but I guess it is optional because it does make the code comments harder to read with the advantage of making the html docs prettier:

```rust
/// Create a new [`TimerFd`](struct.TimerFd.html).
///
/// The timer is initially disarmed and must be armed by calling [`reset`](struct.TimerFd.html#method.reset).
```

P.S. I haven't checked that the documentation links are correct.

src/timerfd.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+impl IntoRawFd for TimerFd {
+    fn into_raw_fd(self) -> RawFd {
+        self.0.into_raw_fd()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
andreeaflorescu: Please prefix tests with `test_` (`test_one_shot`). This is a good way to visually separate tests from helper methods.

src/timerfd.rs
liujing2: Let me split the summary and description. The link might be optional and I found not all of the linux man is appended in original codes.

src/timerfd.rs
liujing2: OK, I will add the prefix for every tests in these files.

src/timerfd.rs
andreeaflorescu: Hmm, I don't think I understand, I meant inside links to structures and functions that you are referring to in the comments.
https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link

src/lib.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+extern crate libc;
+
+#[macro_use]
+pub mod ioctl;
+
+mod errno;
```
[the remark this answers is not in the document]
liujing2: Fixed.

src/eventfd.rs
```diff
@@ -0,0 +1,140 @@
...
+use std::fs::File;
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+use std::{io, mem, result};
+
+use libc::{c_void, dup, eventfd, read, write, EFD_NONBLOCK};
+
+/// A safe wrapper around a Linux eventfd (man 2 eventfd).
+///
+/// An eventfd is useful because it is sendable across processes and can be used for signaling in
+/// and out of the KVM API. They can also be polled like any other file descriptor.
```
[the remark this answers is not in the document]
liujing2: Removed.

Cargo.lock
```diff
@@ -0,0 +1,16 @@
...
+
+[[package]]
+name = "libc-utils"
+version = "0.1.0"
+dependencies = [
+ "libc 0.2.49 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[metadata]
+"checksum libc 0.2.49 (registry+https://github.com/rust-lang/crates.io-index)" = "413f3dfc802c5dc91dc570b05125b6cda9855edfaa9825c9849807876376e70e"
```
jiangliu: There was a discussion about whether Cargo.lock should be committed. I remember the conclusion is not to commit for library crate and only commit for binary crate.

Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "libc-utils"
+version = "0.1.0"
+authors = ["Jing Liu <jing2.liu@linux.intel.com>"]
+edition = "2018"
```
jiangliu: Seems a little risky to force 2018 edition.

src/errno.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
jiangliu: Now it's suggested to dual license for Apache-2.0 and MIT. Please help to add support of MIT.

src/errno.rs
```diff
@@ -0,0 +1,86 @@
...
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Error::new(e.raw_os_error().unwrap_or_default())
+    }
+}
+
+/// Returns the last `errno` as a [`Result`] that is always an error.
+///
+/// [`Result`]: type.Result.html
+///
```
jiangliu: Should we remove the blank lines?

liujing2:
> Could you please be more verbose in each commit message? We should be able to quickly understand what was introduced by each commit only looking at the commit messages IMO.

Sure. Let me append more.

src/errno.rs
liujing2: I think blank line is no need here. The same as others.

Cargo.toml
liujing2: Got it. Remove it and add license BTW.

Cargo.lock
liujing2: Will ignore it in next update.

src/timerfd.rs
liujing2: @andreeaflorescu Thanks for the guide. I was misunderstood. I'll update a new version today.

src/errno.rs
liujing2: > Now it's suggested to dual license for Apache-2.0 and MIT. Please help to add support of MIT.

@jiangliu  I just updated the codes regarding to the license and other review comments. 
BTW, for the vhost utility, could you explain your request in details so I can help?

jiangliu:
Currently the vhost crate makes use of the EventFd and ioctl macros from the sys_util crate, seems they are available in the vmm-sys-util already.

liujing2:
Codes updated with license MIT and review comments.  Thanks!

src/errno.rs
```diff
@@ -0,0 +1,80 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+// SPDX-License-Identifier: dual licence MIT or Apache-2.0
+//
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
```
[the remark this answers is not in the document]
liujing2: BSD license file is added.

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: rust
+
+rust:
+  - stable
+  - nightly
+
+script:
+  - cargo build -- release
+  - cargo test
+  - cargo fmt --all -- --check
```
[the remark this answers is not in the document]
liujing2: Yep, a separate commit on top is all about solving the clippy issue. Thanks.

sboeuf:
@zachreizner @andreeaflorescu @rbradford PTAL

LICENSE.MIT
```diff
@@ -0,0 +1,21 @@
+MIT License
```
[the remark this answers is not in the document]
liujing2: It seems currently dual license for Apache-2.0 and MIT are supported for rust-vmm. See the [comments](https://github.com/rust-vmm/vmm-sys-util/pull/1#discussion_r275105186`) BSD is copied from crosvm.

LICENSE.BSD
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
```
sameo: Could we please rename that file LICENSE-BSD-3-Clause and the Apache one LICENSE-APACHE? Just to be consistent with the already existing crates?

LICENSE.MIT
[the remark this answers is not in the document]
sameo: @liujing2 I think it would be wiser to remove the MIT license for now and only keep the Apache 2.0 and the 3 clause BSD one.

sameo:
@liujing2 I sent you a PR for adding a few more utilities to that crate, mostly for supporting disk image format crates like e.g. QCOW.
See https://github.com/liujing2/vmm-sys-util/pull/1

LICENSE.BSD
liujing2: Will do that.

LICENSE.MIT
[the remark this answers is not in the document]
liujing2: Will remove MIT license.

liujing2:
> @liujing2 I sent you a PR for adding a few more utilities to that crate, mostly for supporting disk image format crates like e.g. QCOW.
> See [liujing2#1](https://github.com/liujing2/vmm-sys-util/pull/1)

Sure, will take a look.

jiangliu:
Seems file_traits, wrtite_zeros, seek_hole and fallocate are related to each other, how about combine them into one file?

jiangliu:
It fails to compile with "cargo build --target x86_64-unknown-linux-musl" as below. GNU library is ok.
error[E0432]: unresolved imports `libc::SEEK_DATA`, `libc::SEEK_HOLE`                            
  --> /usr/local/rust/git/checkouts/vmm-sys-util-5509361781321914/766db44/src/seek_hole.rs:11:28 
   |                                                                                             
11 | use libc::{lseek64, ENXIO, SEEK_DATA, SEEK_HOLE};                                           
   |                            ^^^^^^^^^  ^^^^^^^^^ no `SEEK_HOLE` in the root                  
   |                            |                                                                
   |                            no `SEEK_DATA` in the root. Did you mean to use `SEEK_SET`?      
                                                                                                 
error: aborting due to previous error                                                            
                                                                                                 
For more information about this error, try `rustc --explain E0432`.

liujing2:
> It fails to compile with "cargo build --target x86_64-unknown-linux-musl" as below. GNU library is ok.
> error[E0432]: unresolved imports `libc::SEEK_DATA`, `libc::SEEK_HOLE`
> --> /usr/local/rust/git/checkouts/vmm-sys-util-5509361781321914/766db44/src/seek_hole.rs:11:28
> |
> 11 | use libc::{lseek64, ENXIO, SEEK_DATA, SEEK_HOLE};
> | ^^^^^^^^^ ^^^^^^^^^ no `SEEK_HOLE` in the root
> | |
> | no `SEEK_DATA` in the root. Did you mean to use `SEEK_SET`?

@jiangliu According to the [MUSL libc](https://www.musl-libc.org/download.html) which is used by [rust musl build](https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.html) we can find that it doesn't have SEEK_HOLE and SEEK_HOLE support.

sboeuf:
@zachreizner @dgreid PTAL

jiangliu:
Could we also reexport libc::{EFD_SEMAPHORE, EFD_SEMAPHORE, EFD_CLOEXEC} in the eventfd.rs, so it get a little easier for use.

sameo:
Testing a potential fix...

sameo:
> Testing a potential fix...

Seems to be working.

sameo:
> Could we also reexport libc::{EFD_SEMAPHORE, EFD_SEMAPHORE, EFD_CLOEXEC} in the eventfd.rs, so it get a little easier for use.

@jiangliu Do you mind if we merge that one first and then send that as a follow up PR?

jiangliu:
> > Could we also reexport libc::{EFD_SEMAPHORE, EFD_SEMAPHORE, EFD_CLOEXEC} in the eventfd.rs, so it get a little easier for use.
> 
> @jiangliu Do you mind if we merge that one first and then send that as a follow up PR?

sure, actually I have three patches upon this patch set:)

sameo:
> > > Could we also reexport libc::{EFD_SEMAPHORE, EFD_SEMAPHORE, EFD_CLOEXEC} in the eventfd.rs, so it get a little easier for use.
> > 
> > 
> > @jiangliu Do you mind if we merge that one first and then send that as a follow up PR?
> 
> sure, actually I have three patches upon this patch set:)

Excellent :) Could you please give this PR another review? Adding new commits dismissed the previous reviews...

sboeuf:
Looks like this PR is ready to be merged :)

jiangliu:
> Looks like this PR is ready to be merged :)

Great, the vhost crate could move on once this has been merged:)

PR #3, opened 2019-05-29
Add more utilities to the crate

jiangliu:
Add more utilities to the vmm-sys-util crate:
1) import sock_ctrl_msg.rs from crosvm, which will be used by the vhost crate.
2) import struct_util.rs from crosvm, which is needed by firecracker.
3) implement a safe wrapper over Linux AIO API.
4) Reexport EventFd flags.


src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub mod file_traits;
 pub mod seek_hole;
 pub mod signal;
 pub mod sock_ctrl_msg;
+pub mod struct_util;
```
sameo: As far as I know, this is only needed by the Firecracker kernel loader crate. So this is a case of exporting an API that only has one consumer.
I would prefer that Firecracker either a) define the `struct_util` API in its own kernel loader or b) switch to the rust-vmm `linux-loader` crate. But I don't think we should add it to this rust-vmm crate unless there is more than one user.

jiangliu: I'm OK with both solutions:)

sameo: @jiangliu Sweet :)

Could you also fix the unit test and clippy warnings?

jiangliu: I'm trying to access the buildbike logs:)

sameo: > I'm trying to access the buildbike logs:)

You should be able to see it now.

andreeaflorescu: @sameo it seems like struct_util is also used by linux-loader. Only that instead of using it from vmm-sys-utils the utilities are implemented directly in the linux-loader crate.

I checked the Firecracker code and the only place where we are using the struct_util is in the kernel-loader so if you think it makes more sense there and no other crate can benefit from it we can also just leave it there.

sameo: @andreeaflorescu I think it makes more sense to keep it in the single crate that consumes it, yes.

src/aio.rs
```diff
@@ -0,0 +1,265 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
```
liujing2: Just nit: It seems that the commit message needs to be modified as "... may be used by some ..."

src/sock_ctrl_msg.rs
```diff
@@ -0,0 +1,494 @@
...
+
+enum CmsgBuffer {
+    Inline([u64; (CMSG_BUFFER_INLINE_CAPACITY + 7) / 8]),
+    Heap(Box<[cmsghdr]>),
+}
+
+impl CmsgBuffer {
+    fn with_capacity(capacity: usize) -> CmsgBuffer {
+        let cap_in_cmsghdr_units =
+            (capacity.checked_add(size_of::<cmsghdr>()).unwrap() - 1) / size_of::<cmsghdr>();
```
liujing2: Do we need check the checked_add result here instead of unwrap()?

jiangliu: We assume here that capacity.check_add(size_of::<cmsghdr>()) should not overflow for most cases, otherwise the caller will need to handle error results.

src/aio.rs
```diff
@@ -0,0 +1,265 @@
...
+pub const IOCB_CMD_PREAD: u32 = 0;
+pub const IOCB_CMD_PWRITE: u32 = 1;
+pub const IOCB_CMD_FSYNC: u32 = 2;
+pub const IOCB_CMD_FDSYNC: u32 = 3;
+pub const IOCB_CMD_NOOP: u32 = 6;
+pub const IOCB_CMD_PREADV: u32 = 7;
+pub const IOCB_CMD_PWRITEV: u32 = 8;
+pub const IOCB_FLAG_RESFD: u32 = 1;
+
+/// Limit maximum number of concurrent requests.
```
andreeaflorescu: nit: this should describe what the variable is and not what it is used for.
Suggestion: Maximum number of concurrent requests.

src/aio.rs
```diff
@@ -0,0 +1,265 @@
...
+use std::os::raw::{c_int, c_long, c_uint, c_ulong};
+use std::ptr::null_mut;
+
+type __s16 = ::std::os::raw::c_short;
+type __u16 = ::std::os::raw::c_ushort;
+type __u32 = ::std::os::raw::c_uint;
+type __s64 = ::std::os::raw::c_longlong;
+type __u64 = ::std::os::raw::c_ulonglong;
+
+pub const IOCB_CMD_PREAD: u32 = 0;
```
andreeaflorescu: It's weird that the tests pass even though these pub constants are not documented. Any chance we are missing an `allow(missing_docs)` somewhere?

src/aio.rs
```diff
@@ -0,0 +1,265 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Safe wrapper for Linux AIO API.
+
+#![allow(non_camel_case_types)]
+
+/* automatically generated by rust-bindgen and then manually edited */
```
andreeaflorescu: Please specify version/commit from which you generated the file and which header file was used for generating it.

src/aio.rs
```diff
@@ -0,0 +1,265 @@
...
+pub const IOCB_CMD_FDSYNC: u32 = 3;
+pub const IOCB_CMD_NOOP: u32 = 6;
+pub const IOCB_CMD_PREADV: u32 = 7;
+pub const IOCB_CMD_PWRITEV: u32 = 8;
+pub const IOCB_FLAG_RESFD: u32 = 1;
+
+/// Limit maximum number of concurrent requests.
+pub const MAX_REQUESTS: usize = 0x10000;
+
+/// Rust version of iocb
```
andreeaflorescu: nit: I would write the comment as: 
```
\\\ Wrapper over the [`iocb`](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/aio_abi.h#L73) structure.

File: src/aio.rs [hunk 8]
```diff
@@ -0,0 +1,265 @@
...
+    pub aio_fildes: __u32,
+    pub aio_buf: __u64,
+    pub aio_nbytes: __u64,
+    pub aio_offset: __s64,
+    pub aio_reserved2: __u64,
+    pub aio_flags: __u32,
+    pub aio_resfd: __u32,
+}
+
+/// Rust version of io_event_t.
```
andreeaflorescu: Same comment as for `IoControlBlock`. Also, isn't this a wrapper over `io_event`?
https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/aio_abi.h#L60

src/aio.rs
```diff
@@ -0,0 +1,265 @@
...
+#[repr(C)]
+#[derive(Debug, Default, Copy, Clone)]
+pub struct IoEvent {
+    pub data: __u64,
+    pub obj: __u64,
+    pub res: __s64,
+    pub res2: __s64,
+}
+
+/// Newtype for aio_context_t.
```
andreeaflorescu: Wrapper over `aio_context_t`. Also add link to kernel structure.

src/aio.rs
```diff
@@ -0,0 +1,265 @@
...
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct IoContext(::std::os::raw::c_ulong);
```
andreeaflorescu: I don't know in detail how does the IoContext work, so it might be a dumb question but: is the inner value private on purpose?

src/eventfd.rs
```diff
@@ -11,6 +11,8 @@ use std::{io, mem, result};
 
 use libc::{c_void, dup, eventfd, read, write};
 
+pub use libc::{EFD_CLOEXEC, EFD_NONBLOCK, EFD_SEMAPHORE};
```
andreeaflorescu: nit: commit message summary is too long.
https://chris.beams.io/posts/git-commit/#limit-50

src/sock_ctrl_msg.rs
```diff
@@ -0,0 +1,461 @@
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+//! Used to send and receive messages with file descriptors on sockets that accept control messages
```
andreeaflorescu: nit: Wrappers for sending and receiving ....

We should use nouns for what the module is.

src/aio.rs
jiangliu: You are right, there's no "deny(missing_docs)" in lib.rs.

src/aio.rs
jiangliu: Yes, it's private on purpose, so it could only be accessed through exported public interfaces.

src/aio.rs
jiangliu: Yes, it's wrapper for io_event. Updated the comment:)

src/aio.rs
```diff
@@ -0,0 +1,278 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Safe wrapper for Linux AIO API.
+
+#![allow(non_camel_case_types)]
+
+/* automatically generated by rust-bindgen from file linux/include/uapi/linux/aio_abi.h
+ * of commit 69973b830859 and then manually edited */
```
liujing2: Nit: I think we can only copy 7 figures of commit id here.

src/sock_ctrl_msg.rs
```diff
@@ -0,0 +1,494 @@
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
liujing2: Do we need change "LICENSE" to the real license file name?

jiangliu: yes, I think so

andreeaflorescu:
@jiangliu can you update this PR so we can merge it?

src/aio.rs
```diff
@@ -0,0 +1,325 @@
...
+                aio_nbytes: buf.len() as u64,
+                ..Default::default()
+            },
+        ];
+
+        let mut rc = ctx.submit(&iocbs[..]).unwrap();
+        assert_eq!(rc, 2);
+
+        let mut events = [unsafe { std::mem::uninitialized::<IoEvent>() }];
+        rc = ctx.get_events(1, &mut events, None).unwrap();
```
liujing2: Not sure how can we ensure the event is in completion queue?

src/lib.rs
```diff
@@ -20,9 +20,6 @@ pub mod poll;
 pub mod rand;
 pub mod seek_hole;
 pub mod signal;
-// The rust musl libc implementation has defined msghdr.{__pad1, __pad2) as private, which causes
```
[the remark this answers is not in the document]
jiangliu: The original sock_ctrl_msg.rs is copied from the crosvm project without modification. Is it preferred to separate changes for musl from the original import commit?

src/aio.rs
jiangliu: we are reading /dev/zero, it should not block at all.

src/eventfd.rs
```diff
@@ -30,9 +34,9 @@ impl EventFd {
     ///
     /// ```
     /// extern crate vmm_sys_util;
-    /// use vmm_sys_util::eventfd::EventFd;
+    /// use vmm_sys_util::eventfd::*;
```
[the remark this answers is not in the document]
jiangliu: Fixed.

src/sock_ctrl_msg.rs
```diff
@@ -0,0 +1,461 @@
...
+impl ScmSocket for UnixStream {
+    fn socket_fd(&self) -> RawFd {
+        self.as_raw_fd()
+    }
+}
+
+/// Trait for types that can be converted into an `iovec` that can be referenced by a syscall for
+/// the lifetime of this object.
+///
+/// This trait is unsafe because interfaces that use this trait depend on the base pointer and size
```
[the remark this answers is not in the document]
jiangliu: fixed:)

src/aio.rs
liujing2: It would be nice to copy the test as examples for those functions so that public users can know how to use it explicitly.

src/aio.rs
```diff
@@ -0,0 +1,325 @@
...
+        assert_eq!(rc, 1);
+
+        let mut result = Default::default();
+        let err = ctx
+            .cancel(iocbs[0], &mut result)
+            .unwrap_err()
+            .raw_os_error()
+            .unwrap();
+        assert_eq!(err, libc::EINVAL);
+
```
liujing2: So here we tested a failure cancel() operation right? So do we need check `result`? Meanwhile I am thinking if we need test a successful one?

jiangliu: If you read linux kernel implementation, you will find that aio::cancel() won't success at all:(
Aio is a wired subsystem, and seems no plan to fix it at all.

liujing2: Thanks for the explanation. So we might comment a few words in cancel() implementation, WDYT? And others seems good to me, besides the comment of copying some Examples. :)

src/lib.rs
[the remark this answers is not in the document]
andreeaflorescu: I didn't realize it. You can leave it in separate commits in this case.

nit: can you also specify the CrosVM commit in your commit message. It might be useful in the future.

src/aio.rs
jiangliu: Fixed:)

src/lib.rs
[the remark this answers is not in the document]
jiangliu: done:)

src/aio.rs
jiangliu: done:)

PR #4, opened 2019-05-31
define wrapper for C FAM structs

serban300:
In C an array of unknown size may appear within a struct definition
as the last member (as long as there is at least one other named member).
This is a known as flexible array member.

Pre C99, the same behavior could be achieved using zero length arrays.

Flexible Array Members are the go-to choice for working with large
amounts of data prefixed by header values.

KVM uses this type of structures (e.g., kvm_cpuid2, kvm_msr_list, etc.)

FamStructWrapper provides a common abstraction for this type of
structures and helps in treating them similarly to Rust Vec.

### DIFF
--- a/src/fam.rs
+++ b/src/fam.rs
@@ -1,0 +1,696 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
+
+use std::mem::{self, size_of};
+
+/// Errors associated with the [`FamStructWrapper`](struct.FamStructWrapper.html) struct.
+#[derive(Clone, Debug)]
+pub enum Error {
+    /// The max size has been exceeded
+    SizeLimitExceeded,
+}
+
+/// Trait for accessing some properties of C defined FAM structures.
+///
+/// In C 99 an array of unknown size may appear within a struct definition as the last member
+/// (as long as there is at least one other named member).
+/// This is known as a flexible array member (FAM).
+/// Pre C99, the same behavior could be achieved using zero length arrays.
+///
+/// Flexible Array Members are the go-to choice for working with large amounts of data
+/// prefixed by header values.
+///
+/// For example the kvm API has many structures that are very similar to the `MockFamStruct`
+/// structure defined in the example bellow.
+///
+/// This is unsafe due to the number of constraints that aren't checked:
+/// * the implementer should be a POD
+/// * the implementor should contain a flexible array member of elements of type `Entry`
+/// * `Entry` should be a POD
+///
+/// Violating these may cause problems.
+///
+/// # Example
+///
+/// ```
+/// use vmm_sys_util::fam::*;
+///
+/// #[repr(C)]
+/// #[derive(Default)]
+/// pub struct __IncompleteArrayField<T>(::std::marker::PhantomData<T>, [T; 0]);
+/// impl<T> __IncompleteArrayField<T> {
+///     #[inline]
+///     pub fn new() -> Self {
+///         __IncompleteArrayField(::std::marker::PhantomData, [])
+///     }
+///     #[inline]
+///     pub unsafe fn as_ptr(&self) -> *const T {
+///         ::std::mem::transmute(self)
+///     }
+///     #[inline]
+///     pub unsafe fn as_mut_ptr(&mut self) -> *mut T {
+///         ::std::mem::transmute(self)
+///     }
+///     #[inline]
+///     pub unsafe fn as_slice(&self, len: usize) -> &[T] {
+///         ::std::slice::from_raw_parts(self.as_ptr(), len)
+///     }
+///     #[inline]
+///     pub unsafe fn as_mut_slice(&mut self, len: usize) -> &mut [T] {
+///         ::std::slice::from_raw_parts_mut(self.as_mut_ptr(), len)
+///     }
+/// }
+///
+/// impl<T> ::std::clone::Clone for __IncompleteArrayField<T> {
+///     #[inline]
+///     fn clone(&self) -> Self {
+///         Self::new()
+///     }
+/// }
+///
+/// #[repr(C)]
+/// #[derive(Clone, Default)]
+/// struct MockFamStruct {
+///     pub len: u32,
+///     pub padding: u32,
+///     pub entries: __IncompleteArrayField<u32>,
+/// }
+///
+/// unsafe impl FamStruct for MockFamStruct {
+///     type Entry = u32;
+///
+///     fn len(&self) -> usize {
+///         self.len as usize
+///     }
+///
+///     fn set_len(&mut self, len: usize) {
+///         self.len = len as u32
+///     }
+///
+///     fn max_len() -> usize {
+///         100
+///     }
+///
+///     fn as_slice(&self) -> &[u32] {
+///         let len = self.len();
+///         unsafe { self.entries.as_slice(len) }
+///     }
+///
+///     fn as_mut_slice(&mut self) -> &mut [u32] {
+///         let len = self.len();
+///         unsafe { self.entries.as_mut_slice(len) }
+///     }
+/// }
+///
+/// type MockFamStructWrapper = FamStructWrapper<MockFamStruct>;
+/// ```
+///
+#[allow(clippy::len_without_is_empty)]
+pub unsafe trait FamStruct {
+    /// The type of the FAM entries
+    type Entry: PartialEq + Copy;
+
+    /// Get the FAM length
+    ///
+    /// These type of structures contain a member that holds the FAM length.
+    /// This method will return the value of that member.
+    fn len(&self) -> usize;
+
+    /// Set the FAM length
+    ///
+    /// These type of structures contain a member that holds the FAM length.
+    /// This method will set the value of that member.
+    fn set_len(&mut self, len: usize);
+
+    /// Get max allowed FAM length
+    ///
+    /// This depends on each structure.
+    /// For example a structure representing the cpuid can contain at most 80 entries.
+    fn max_len() -> usize;
+
+    /// Get the FAM entries as slice
+    fn as_slice(&self) -> &[Self::Entry];
+
+    /// Get the FAM entries as mut slice
+    fn as_mut_slice(&mut self) -> &mut [Self::Entry];
+}
+
+/// A wrapper for [`FamStruct`](trait.FamStruct.html)
+///
+/// It helps in treating a [`FamStruct`](trait.FamStruct.html) similarly to an actual `Vec`.
+pub struct FamStructWrapper<T: Default + FamStruct> {
+    // this variable holds the FamStruct structure. We use a `Vec<T>` to make the allocation
+    // large enough while still being aligned for `T`. Only the first element of `Vec<T>`
+    // will actually be used as a `T`. The remaining memory in the `Vec<T>` is for `entries`,
+    // which must be contiguous. Since the entries are of type `FamStruct::Entry` we must
+    // be careful to convert the desired capacity of the `FamStructWrapper`
+    // from `FamStruct::Entry` to `T` when reserving or releasing memory.
+    mem_allocator: Vec<T>,
+}
+
+impl<T: Default + FamStruct> FamStructWrapper<T> {
+    /// Convert FAM len to `mem_allocator` len
+    ///
+    /// Get the capacity required by mem_allocator in order to hold
+    /// the provided number of [`FamStruct::Entry`](trait.FamStruct.html#associatedtype.Entry)
+    fn mem_allocator_len(fam_len: usize) -> usize {
+        let wrapper_size_in_bytes = size_of::<T>() + fam_len * size_of::<T::Entry>();
+        (wrapper_size_in_bytes + size_of::<T>() - 1) / size_of::<T>()
+    }
+
+    /// Convert `mem_allocator` len to FAM len
+    ///
+    /// Get the number of elements of type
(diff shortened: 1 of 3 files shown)


src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,586 @@
...
+
+        FamStructWrapper {
+            mem_allocator,
+            len: num_elements,
+            capacity: num_elements,
+        }
+    }
+
+    /// Creates a new `FamStructWrapper` structure based on a supplied slice
+    /// of `FamStruct::Entry`.
```
aghecenco:
```suggestion
    /// of [`FamStruct::Entry`](trait.FamStruct.html#associatedtype.Entry).
```

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,586 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `entries` - The vector of `FamStruct::Entry` entries.
```
aghecenco:
```suggestion
    /// * `entries` - The vector of [`FamStruct::Entry`](trait.FamStruct.html#associatedtype.Entry) entries.
```

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,586 @@
...
+
+        {
+            let wrapper_entries = adapter.as_mut_fam_struct().as_mut_slice();
+            wrapper_entries.copy_from_slice(entries);
+        }
+
+        adapter
+    }
+
+    /// Get a reference to the actual `FamStruct` instance.
```
aghecenco:
```suggestion
    /// Get a reference to the actual [`FamStruct`](trait.FamStruct.html) instance.
```

This applies to all the following mentions too, I'll stop with the suggestions :smile:

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,586 @@
...
+    ///                    `FamStruct::Entry` in the initial `FamStructWrapper`
+    ///
+    pub fn new(num_elements: usize) -> FamStructWrapper<T> {
+        let required_mem_allocator_capacity =
+            FamStructWrapper::<T>::mem_allocator_len(num_elements);
+
+        let mut mem_allocator = Vec::with_capacity(required_mem_allocator_capacity);
+        for _ in 0..required_mem_allocator_capacity {
+            mem_allocator.push(T::default())
+        }
```
jiangliu: Should we initialize memory other than first element to zero instead of T::default()? T::default() may generate some non-zero content.

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,586 @@
...
+        mem_allocator[0].set_len(num_elements);
+
+        FamStructWrapper {
+            mem_allocator,
+            len: num_elements,
+            capacity: num_elements,
+        }
+    }
+
+    /// Creates a new `FamStructWrapper` structure based on a supplied slice
```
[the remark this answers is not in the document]
serban300: Fixed

src/fam_struct_wrapper.rs
serban300: Fixed

src/fam_struct_wrapper.rs
serban300: Fixed

src/fam_struct_wrapper.rs
serban300: Fixed

andreeaflorescu:
@alexandruag I know you had some comments on this PR. Can you please add them here in case they weren't already fixed?

I will also check if the comments I had were resolved.

serban300:
> FYI I can't get coverage with `cargo kcov`, could you?
> 
> ```shell
> test fam_struct_wrapper::tests::test_reserve ... ok
> kcov: Process exited with signal 4 (SIGILL) at 0x7ffff6ce00e3
> 
> kcov: Illegal instructions are sometimes caused by some GCC versions
> kcov: miscompiling C++ headers. If the problem is persistent, try running
> kcov: with --verify. For more information, see
> kcov: http://github.com/SimonKagstrom/kcov/issues/18
> ```

I forgot about this. I will try it.

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,590 @@
...
+        }
+
+        // This is safe since this method is not increasing the len
+        self.set_len(num_kept_entries).expect("invalid length");
+    }
+}
+
+impl<T: Default + FamStruct> PartialEq for FamStructWrapper<T> {
+    fn eq(&self, other: &FamStructWrapper<T>) -> bool {
+        self.len == other.len && self.as_slice() == other.as_slice()
```
alexandruag: Sorry I know I asked this regarding the old PR as well, but I don't remember the conclusion :( Is it a valid occurence to have `self.as_slice() == other.as_slice()`, but `self.len != other.len`?

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,590 @@
...
+    }
+
+    /// Updates the length of `self` to the specified value.
+    /// Also updates the length of the `T` structure and of `self.mem_allocator` accordingly.
+    /// Reserves additional capacity if the len is increased.
+    /// Deallocates unnecessary memory if the len is decreased.
+    ///
+    /// # Error: When len is greater than the max possible len it returns Error::SizeLimitExceeded
+    ///
+    fn set_len(&mut self, len: usize) -> Result<(), Error> {
```
alexandruag: It seems that when `set_len` causes the length to increase, the newly added elements are based on basically random memory contents (I don't think reserving capacity also initializes the memory to a particular value). If that's true, is this a part of the `set_len` method contract, or we should try to zero-initialize the new elements like we do for `new`?

src/fam_struct_wrapper.rs
```diff
@@ -0,0 +1,590 @@
...
+/// #[derive(Clone, Default)]
+/// struct MockFamStruct {
+///     pub len: u32,
+///     pub padding: u32,
+///     pub entries: [u32; 0],
+/// }
+/// ```
+///
+#[allow(clippy::len_without_is_empty)]
+pub trait FamStruct {
```
alexandruag: Also, I think we should mark this trait as `unsafe`, just so the user gives an explicit sign-off that the implementation adheres to the conventions established by the trait, including that any type implementing this must be valid to initialize from any random bits. Without this assumption,we can't be sure that zero-initializing the memory in `new` and using `Vec::set_len` are actually safe to do in the implementation of `FamStructWrapper`. What do you ppl think?

src/fam.rs
```diff
@@ -0,0 +1,623 @@
...
+/// #[derive(Clone, Default)]
+/// struct MockFamStruct {
+///     pub len: u32,
+///     pub padding: u32,
+///     pub entries: [u32; 0],
+/// }
+/// ```
+///
+#[allow(clippy::len_without_is_empty)]
+pub unsafe trait FamStruct {
```
alexandruag: Looks good, we should now think what the contract is such that implementing `FamStruct` is actually safe, so we can add it to the documentation. So far, likely mentions are the implementer and `Entry` should essentially be POD (plain old data) structs, and `size_of(Entry)` should be a multiple of `align_of(Entry)`.

src/fam.rs
```diff
@@ -0,0 +1,623 @@
...
+/// This is known as a flexible array member (FAM).
+/// Pre C99, the same behavior could be achieved using zero length arrays.
+///
+/// Flexible Array Members are the go-to choice for working with large amounts of data
+/// prefixed by header values.
+///
+/// For example the kvm API has many structures that are very similar to the following
+/// `MockFamStruct` structure:
+///
+/// # Example
```
andreeaflorescu: This example is not really relevant to the code, but rather an example of a structure with a zero size array. The `Examples` section is used for providing an example of using the code it refers to. For reference: https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments

I would rather delete the example or implement `FamStruct` for `MockFamStruct` so it makes sense.

src/fam.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    /// Updates the length of `self` to the specified value.
+    /// Also updates the length of the `T` structure and of `self.mem_allocator` accordingly.
+    /// Reserves additional capacity if the len is increased.
+    /// Deallocates unnecessary memory if the len is decreased.
+    ///
+    /// This method might trigger reallocations of the underlying buffer.
+    ///
+    /// # Error: When len is greater than the max possible len it returns Error::SizeLimitExceeded
```
andreeaflorescu: This will look a bit weird in markdown. You basically have a heading title that is very very long.

Please use as heading title `Errors` as this is the standard defined in the Rust documentation guidelines (even if the function can return just on type of error). The explanation for the error needs to be in a paragraph.

src/fam.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    impl<T> ::std::clone::Clone for __IncompleteArrayField<T> {
+        #[inline]
+        fn clone(&self) -> Self {
+            Self::new()
+        }
+    }
+
+    #[repr(C)]
```
andreeaflorescu: I would put the following lines (till 398) as a documentation example for the fam module.

src/fam.rs
serban300: I didn't write anything related to: `size_of(Entry) should be a multiple of align_of(Entry)`

We'll sync offline.

jiangliu:
What's the plan with this PR? I saw this pattern has been repeated in several crates, so it may help to reduce common code.

serban300:
> What's the plan with this PR? I saw this pattern has been repeated in several crates, so it may help to reduce common code.

We have to discuss the final details related to [this](https://github.com/rust-vmm/vmm-sys-util/pull/4#discussion_r302034975) comment. I'll try to sync with @alexandruag about this as soon as possible.

alexandruag:
> We have to discuss the final details related to [this](https://github.com/rust-vmm/vmm-sys-util/pull/4#discussion_r302034975) comment. I'll try to sync with @alexandruag about this as soon as possible.

I've had another look and for now there are no other pressing concerns from my point of view. I was worried before there were dangerous corner cases when the size of `Entry` is not a multiple of its alignment, but it looks like the compiler ensures that's never the case via padding bytes when necessary.

src/fam.rs
```diff
@@ -0,0 +1,695 @@
...
+///         let len = self.len();
+///         unsafe { self.entries.as_mut_slice(len) }
+///     }
+/// }
+///
+/// type MockFamStructWrapper = FamStructWrapper<MockFamStruct>;
+/// ```
+///
+#[allow(clippy::len_without_is_empty)]
+pub unsafe trait FamStruct {
```
liujing2: Could you please add some documentation on this trait? Some references here https://github.com/rust-vmm/vmm-sys-util/blob/9014b7ab94461b54b90ca64c6d8a442b6b4b84c8/src/terminal.rs#L72

src/fam.rs
```diff
@@ -0,0 +1,695 @@
...
+    /// Get the array entries as slice
+    fn as_slice(&self) -> &[Self::Entry];
+
+    /// Get the array entries as mut slice
+    fn as_mut_slice(&mut self) -> &mut [Self::Entry];
+}
+
+/// A wrapper that helps in treating a [`FamStruct`](trait.FamStruct.html) similarly to
+/// an actual `Vec`.
+///
```
liujing2: Remove this line if no other description.

liujing2: Also, could you please kindly separate the documentation into a short summary line of its functionality on the first line, and more explanation in a new paragraph? Ditto for others below.

src/fam.rs
```diff
@@ -0,0 +1,695 @@
...
+    /// Constructs a new [`FamStructWrapper<T>`](struct.FamStructWrapper.html) that contains
+    /// `num_elements` zero-initialized elements of type
+    /// [`FamStruct::Entry`](trait.FamStruct.html#associatedtype.Entry)
+    ///
+    /// # Arguments
+    ///
+    /// * `num_elements` - The number of zero-initialized elements of type
+    ///                    [`FamStruct::Entry`](trait.FamStruct.html#associatedtype.Entry)
+    ///                    in the initial [`FamStructWrapper`](struct.FamStructWrapper.html)
+    ///
```
liujing2: Yep, good to see the detailed #Arguments description. :)

src/fam.rs
```diff
@@ -0,0 +1,695 @@
...
+        assert_eq!(adapter.len(), desired_len);
+        for element in adapter.as_slice() {
+            assert_eq!(*element, 0 as u32);
+        }
+
+        // decrease len
+        desired_len = 5;
+        assert!(adapter.set_len(desired_len).is_ok());
+        assert_eq!(adapter.len(), desired_len);
+        // check that the capacity
```
liujing2: Useless comment?

src/fam.rs
serban300: Done

src/fam.rs
serban300: I added just a couple of lines. I couldn't think of any other details that would need to be specified. Please let me know if anything is not clear.

src/fam.rs
```diff
@@ -0,0 +1,696 @@
...
+///     }
+///     #[inline]
+///     pub unsafe fn as_slice(&self, len: usize) -> &[T] {
+///         ::std::slice::from_raw_parts(self.as_ptr(), len)
+///     }
+///     #[inline]
+///     pub unsafe fn as_mut_slice(&mut self, len: usize) -> &mut [T] {
+///         ::std::slice::from_raw_parts_mut(self.as_mut_ptr(), len)
+///     }
+/// }
```
jiangliu: Could the __IncompleteArrayField struct be exported as a pub generic type, seems it only contains generic code.

serban300: This struct is generated automatically by bindgen. You can find it for example in the kvm-bindings project. I needed to copy it just for the unit tests and the example. I don't know if it adds any value exporting it.

Issue #6, opened 2019-06-04
Missing integration tests


andreeaflorescu:
I just noticed that vmm-sys-utils is not testing the musl build. There are also some other tests missing:
- build with musl on x86 and arm
- unit tests with musl on x86 and arm
- warning checks on x86 and arm

For examples you can check the [kvm-ioctls pipeline](https://github.com/rust-vmm/kvm-ioctls/blob/master/.buildkite/pipeline.yml).

Issue #5, opened 2019-05-31
Use container from rustvmm org


andreeaflorescu:
Replace fandree/rust-vmm-dev with rustvmm/dev:v1 in Buildkite pipeline.

dev:v1 has exactly the same packages and rust version as fandree/rust-vmm-dev so it should be a seamless change.

sameo:
Fixed with #8

PR #8, opened 2019-06-04
musl: adjust for incompatibility between standard library and libc

### LINKED ISSUE #7: Build & tests fail with musl target
Tried to build vmm-sys-utils with musl and it fails:

```cargo test --all --target x86_64-unknown-linux-musl
   Compiling libc v0.2.58
   Compiling vmm-sys-util v0.1.0 (/home/local/ANT/fandree/sources/work/rust-vmm/vmm-sys-util)
error[E0308]: mismatched types
   --> src/signal.rs:317:9
    |
316 |     fn pthread_handle(&self) -> pthread_t {
    |                                 --------- expected `*mut libc::c_void` because of return type
317 |         self.as_pthread_t()
    |         ^^^^^^^^^^^^^^^^^^^ expected *-ptr, found u64
    |
    = note: expected type `*mut libc::c_void`
               found type `u64`

error: aborting due to previous error
```

bonzini:
This should fix #7.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,7 +8,7 @@ steps:
       platform: x86_64.metal
     plugins:
       - docker#v3.0.1:
-          image: "fandree/rust-vmm-dev"
+          image: "rustvmm/dev:v2"
           always-pull: true
 
   - label: "build-gnu-arm"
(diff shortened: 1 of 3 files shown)


andreeaflorescu:
Can you also add the musl tests to the pipeline to make sure the issue is actually fixed?

bonzini:
Will also fix #5 since I was at it.

bonzini:
And will fix #6. :)

.buildkite/pipeline.yml
```diff
@@ -71,5 +111,31 @@ steps:
       platform: x86_64.metal
```
andreeaflorescu: Since you are already using v2, can you also add the clippy test for aarch64?
```
  - label: "clippy-arm"
      commands:
       - cargo clippy --all
      retry:
        automatic: false
      agents:
        platform: arm.metal
      plugins:
        - docker#v3.0.1:
            image: "rustvmm/dev:v2"
            always-pull: true
```

I was just doing that for kvm-ioctls as well. Starting with Rust 1.34 clippy is available on aarch64 as well.

bonzini:
@sameo Can you review?

Issue #10, opened 2019-07-01
Add deny(missing_docs) check to src/lib.rs


jiangliu:
All crates for rust-vmm should enable #![deny(missing_docs)] check in src/lib.rs.

PR #12, opened 2019-07-17
Prepare documents for publishing

liujing2:
Two patches for license issue and documents issue. Fix https://github.com/rust-vmm/vmm-sys-util/issues/10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 // Copyright 2019 Intel Corporation. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
 extern crate libc;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,31 +1,41 @@
 // Copyright 2019 Intel Corporation. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
+//! `vmm-sys-util` is a collection of utilites to make vmm implementation
+//! more convenient.
+
+#![deny(missing_docs)]
+
 extern crate libc;
 
 mod tempdir;
 
 #[macro_use]
-pub mod ioctl;
+mod ioctl;
 
-pub mod errno;
-pub mod eventfd;
-pub mod file_traits;
-pub mod seek_hole;
-pub mod signal;
-pub mod terminal;
-pub mod timerfd;
-pub mod write_zeroes;
+mod errno;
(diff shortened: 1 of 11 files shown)


liujing2:
@sameo @andreeaflorescu @jiangliu Thanks.

src/lib.rs
```diff
@@ -1,40 +1,55 @@
...
+
+#![deny(missing_docs)]
+
 extern crate libc;
 
 mod tempdir;
 
 #[macro_use]
-pub mod ioctl;
+mod ioctl;
```
andreeaflorescu: I prefer having the functionality exported with modules because in this case the modules actually have the purpose of doing a logical separation of components. I know that implies documenting the modules as well, but it shouldn't be super hard to do it.

src/syslog.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
andreeaflorescu: This is not needed.

src/syslog.rs
```diff
@@ -86,6 +88,7 @@ impl fmt::Display for Priority {
 /// The facility of a syslog message.
 ///
 /// See syslog man pages for information on their semantics.
+#[allow(missing_docs)]
```
andreeaflorescu: I find facility to be a really terrible name. I guess we keep it because it's the same used by `syslog`, it sort of makes sense. But I would still be more descriptive in the documentation.

src/lib.rs
```diff
@@ -1,40 +1,55 @@
...
 use std::os::unix::io::AsRawFd;
 
 pub use crate::file_traits::{FileSetLen, FileSync};
 pub use crate::seek_hole::SeekHole;
 pub use crate::write_zeroes::{PunchHole, WriteZeroes};
 
+/// The mode list used by [`fallocate`]
+///
+/// [`fallocate`]: fn.fallocate.html
 pub enum FallocateMode {
```
andreeaflorescu: Why is `FallocateMode` placed in the root of vmm-sys-utils? Should we have it in one of the existing modules or create a module for this functionality?

andreeaflorescu: I fixed this in my PR: https://github.com/rust-vmm/vmm-sys-util/pull/14

src/lib.rs
liujing2: Agreed. Public modules would be good than directly publishing APIs.

liujing2:
I'd like to rebase the PR on #14 after it is merged. And adding more improvements.

liujing2:
Let's prepare the documentation in separate PR later. Let me close it for now.

PR #14, opened 2019-07-17
Small improvements to public documentation

andreeaflorescu:
I only updated the documentation for the errno module by adding some more details to existing docs and  examples.
I also did a bit of cleanup in the way we are exporting functionality from this crate. The major change is having all functionality exported in public modules as oppose to export some of the functionality as bulk using `pub use mod_name::*`. This change also improves the readability of the documentation. Check it out with ```cargo doc --open```.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -22,7 +22,7 @@ use libc::{
     EPOLL_CLOEXEC, EPOLL_CTL_ADD, EPOLL_CTL_DEL, EPOLL_CTL_MOD,
 };
 
-use crate::{errno_result, Error, Result};
+use crate::errno::{errno_result, Error, Result};
 
 macro_rules! handle_eintr_errno {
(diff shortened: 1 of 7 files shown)


src/poll.rs
```diff
@@ -463,8 +463,10 @@ impl<T: PollToken> IntoRawFd for EpollContext<T> {
 /// # Example
 ///
 /// ```
-/// # use vmm_sys_util::{Result, EventFd, PollContext, PollEvents};
-/// # fn test() -> Result<()> {
+/// # use vmm_sys_util::errno::Result;
+/// # use vmm_sys_util::eventfd::EventFd;
+/// # use vmm_sys_util::poll::{PollContext, PollEvents};
+/// fn test() -> Result<()> {
```
liujing2: I like this changes. This makes more sense than before. :)

PR #16, opened 2019-07-23
rand: support getting random numbers and alphanumeric strings

forbesmyester:
Signed-off-by: Matt Forrester <github.com@speechmarks.com>

To implement the `tempfile` replacement I need the ability to generate random strings, thus random numbers also. This is taken from firecracker ( https://github.com/firecracker-microvm/firecracker/blob/master/fc_util/src/lib.rs ) as well as the alphanumeric stuff by myself.

I felt the need to add an author log for credits / possible issue tracking as we'll loose that across the repo's

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -1,0 +1,129 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Miscellaneous functions related to getting (pseudo) random numbers and
+//! strings.
+//!
+//! NOTE! This should not be used when you do need __real__ random numbers such
+//! as for encryption but will probably be suitable when you want locally
+//! unique ID's that will not be shared over the network.
+
+extern crate libc;
+use std::ffi::OsString;
+use std::str;
+
+/// Gets an ever increasing u64 (at least for this process).
+///
+/// The number retreived will be based upon the time of the last reboot (x86_64)
+/// and something undefined for other architectures.
+pub fn timestamp_cycles() -> u64 {
+    #[cfg(target_arch = "x86_64")]
+    // Safe because there's nothing that can go wrong with this call.
+    unsafe {
+        std::arch::x86_64::_rdtsc() as u64
+    }
+
+    #[cfg(not(target_arch = "x86_64"))]
+    {
+        const MONOTONIC_CLOCK_MULTPIPLIER: u64 = 1_000_000_000;
+
+        let mut ts = libc::timespec {
+            tv_sec: 0,
+            tv_nsec: 0,
+        };
+
+        unsafe {
+            libc::clock_gettime(libc::CLOCK_MONOTONIC, &mut ts);
+        }
+        (ts.tv_sec as u64) * MONOTONIC_CLOCK_MULTPIPLIER + (ts.tv_nsec as u64)
+    }
+}
+
+/// Generate pseudo random u32 numbers based on the current timestamp.
+pub fn xor_psuedo_rng_u32() -> u32 {
+    let mut t: u32 = timestamp_cycles() as u32;
+    // Taken from https://en.wikipedia.org/wiki/Xorshift
+    t ^= t << 13;
+    t ^= t >> 17;
+    t ^ (t << 5)
+}
+
+// This will get an array of numbers that can safely be converted to strings
+// because they will be in the range [a-zA-Z0-9]. The return vector could be any
+// size between 0 and 4.
+fn xor_psuedo_rng_u8_alphanumerics(rand_fn: &Fn() -> u32) -> Vec<u8> {
+    let mut r = vec![];
+
+    fn between(lower: u8, upper: u8, to_check: u8) -> bool {
+        (to_check >= lower) && (to_check <= upper)
+    }
+
+    for n in &rand_fn().to_ne_bytes() {
+        // Upper / Lower alphabetics and numbers.
+        if between(48, 57, *n) || between(65, 90, *n) || between(97, 122, *n) {
+            r.push(*n);
+        }
+    }
+    r
+}
+
+fn rand_alphanumerics_impl(rand_fn: &Fn() -> u32, len: usize) -> OsString {
+    let mut buf = OsString::new();
+    let mut done = 0;
+    loop {
+        for n in xor_psuedo_rng_u8_alphanumerics(rand_fn) {
+            done += 1;
+            buf.push(str::from_utf8(&[n]).unwrap_or("_"));
+            if done >= len {
+                return buf;
+            }
+        }
+    }
+}
+
+/// Gets a pseudo random OsString of length `len` with characters in the
+/// range [a-zA-Z0-9].
+pub fn rand_alphanumerics(len: usize) -> OsString {
+    rand_alphanumerics_impl(&xor_psuedo_rng_u32, len)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_timestamp_cycles() {
+        for _ in 0..1000 {
+            assert!(timestamp_cycles() < timestamp_cycles());
+        }
+    }
+
+    #[test]
+    fn test_xor_psuedo_rng_u32() {
+        for _ in 0..1000 {
+            assert_ne!(xor_psuedo_rng_u32(), xor_psuedo_rng_u32());
+        }
+    }
+
+    #[test]
+    fn test_xor_psuedo_rng_u8_alphas() {
+        let i = 3612982; // 55 (shifted 16 places), 33 (shifted 8 places), 54...
+                         // The 33 will be discarded as it is not a valid letter
+                         // (upper or lower) or number.
+        let s = xor_psuedo_rng_u8_alphanumerics(&|| i);
+        assert_eq!(vec![54, 55], s);
+    }
+
+    #[test]
+    fn test_rand_alphanumerics_impl() {
+        let s = rand_alphanumerics_impl(&|| 14134, 5);
+        assert_eq!("67676", s);
+    }
+
+    #[test]
+    fn test_rand_alphanumerics() {
+        let s = rand_alphanumerics(5);
+        assert_eq!(5, s.len());
+    }
+
+}
--- a/coverage_config.json
+++ b/coverage_config.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 67.9,
+  "coverage_score": 69.3,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,7 @@ pub mod eventfd;
 pub mod fallocate;
 pub mod file_traits;
 pub mod poll;
+pub mod rand;
 pub mod seek_hole;
 pub mod signal;
 pub mod tempdir;


jiangliu:
Is this util targeting test only? Or also for generic usages?

forbesmyester:
Well...

I'm working on [this](https://github.com/firecracker-microvm/firecracker/issues/1125) issue in the firecracker code base, which is really about removing a large `tempfile` dependency.

The dependency essentially says "create me a random dir/file in directory X" and uses a random set of letters/numbers for part of the name. For that functionality to work it needs to generate random strings.

However, the random (for numbers) functionality already existed in firecracker (something to do with networking) and I have expanded it to simple (safe) strings, as I need that functionality and I __think__ it's probably useful in other places.

I suppose the tempfile/tempdir functionality could be implemented using a counter instead, I'm not sure if that better or not, might cause major slowdown because of IO depending on where it's used. In any case I'm trying to introduce as little change as possible to the current way it's done currently.

So I suppose the answer to your question is it probably is for generic usage as I'm presenting it here. 

If we don't feel the functionality is useful then I need to think again, perhaps using a counter as above? (I'd need to ask the Firecracker team).

If the code needs improvement to bring it up to the required quality I'm happy to do that (I'm very new to Rust and this is a learning exercise for me).

andreeaflorescu:
@jiangliu do you have any major concerns with the code?

src/rand.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Authors (alpha order): Adrian Catangiu @acatangiu, Alexandra Lordache @aghecenco, Alexandru Agache @alexandruag, Diana Popa @dianpopa, Matt Forrester @forbesmyester, Radu Weiss @raduweiss
+
```
andreeaflorescu: The module needs high level documentation which can be added with `!\\\` instead of just `\\\`.
https://doc.rust-lang.org/1.9.0/book/documentation.html#documenting-modules

src/rand.rs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        let mut ts = libc::timespec {
+            tv_sec: 0,
+            tv_nsec: 0,
+        };
+
+        unsafe {
+            libc::clock_gettime(libc::CLOCK_MONOTONIC, &mut ts);
+        }
+        (ts.tv_sec as u64) * 1000000000 + (ts.tv_nsec as u64)
```
andreeaflorescu: Can you add a define here instead of 1000000000? Also, clippy will complain if you use large numbers without formatting. It should probably be changed to `1_000_000_000`.

src/rand.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Authors (alpha order): Adrian Catangiu @acatangiu, Alexandra Lordache @aghecenco, Alexandru Agache @alexandruag, Diana Popa @dianpopa, Matt Forrester @forbesmyester, Radu Weiss @raduweiss
+
+extern crate libc;
+use std::ffi::OsString;
+use std::str;
+
+pub fn timestamp_cycles() -> u64 {
```
andreeaflorescu: Public functions need documentation.

src/rand.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+        (ts.tv_sec as u64) * 1000000000 + (ts.tv_nsec as u64)
+    }
+}
+
+fn timespec_to_us(time_struct: &libc::timespec) -> u64 {
+    (time_struct.tv_sec as u64) * 1_000_000 + (time_struct.tv_nsec as u64) / 1000
+}
+
+pub fn now_cputime_us() -> u64 {
```
andreeaflorescu: This function also needs documentation.

src/rand.rs
```diff
@@ -0,0 +1,136 @@
...
+        tv_sec: 0,
+        tv_nsec: 0,
+    };
+    // Safe because the parameters are valid.
+    unsafe { libc::clock_gettime(libc::CLOCK_PROCESS_CPUTIME_ID, &mut time_struct) };
+    timespec_to_us(&time_struct)
+}
+
+
+// This generates pseudo random u32 numbers based on the current timestamp. Only works for x86_64,
```
andreeaflorescu: This comment can be changed to:
`\\\ Generate pseudo random u32 numbers based on the current timestamp.`
This will change the comment in a documentation comment.

Also, if this works only on x86_64 it should be labelled accordingly with `#[cfg(target_arch = "x86_64")]` and the comment about it only working on x86_64 should be deleted.

jiangliu:
> @jiangliu do you have any major concerns with the code?

It's ok for using by test cases, but not sure whether the entropy source is safe for product code.

src/rand.rs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Authors (alpha order): Adrian Catangiu @acatangiu, Alexandra Lordache @aghecenco, Alexandru Agache @alexandruag, Diana Popa @dianpopa, Matt Forrester @forbesmyester, Radu Weiss @raduweiss
```
andreeaflorescu: We don't typically single out individual contributors in the files as this is very easy to miss and is not going to be updated.

src/rand.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        unsafe {
+            libc::clock_gettime(libc::CLOCK_MONOTONIC, &mut ts);
+        }
+        (ts.tv_sec as u64) * 1000000000 + (ts.tv_nsec as u64)
+    }
+}
+
+fn timespec_to_us(time_struct: &libc::timespec) -> u64 {
+    (time_struct.tv_sec as u64) * 1_000_000 + (time_struct.tv_nsec as u64) / 1000
```
andreeaflorescu: This code could benefit from some comments about what is the logic here.
Can you also add defines for the magic numbers or explain in a comment what they represent?

andreeaflorescu:
@jiangliu we can add your concern as a comment for the functions that have this problem. What do you say? If you would agree with this approach, can you comment on the individual functions where you think the entropy is not appropriate?

jiangliu:
> @jiangliu we can add your concern as a comment for the functions that have this problem. What do you say? If you would agree with this approach, can you comment on the individual functions where you think the entropy is not appropriate?

Great, I'm OK with this solution:)

src/rand.rs
```diff
@@ -0,0 +1,136 @@
...
+
+extern crate libc;
+use std::ffi::OsString;
+use std::str;
+
+pub fn timestamp_cycles() -> u64 {
+    #[cfg(target_arch = "x86_64")]
+    // Safe because there's nothing that can go wrong with this call.
+    unsafe {
+        std::arch::x86_64::_rdtsc() as u64
```
jiangliu: The return value of rdtsc() is some sort of predictable, so it should not be used for cryptographic.

src/rand.rs
forbesmyester: Yep

src/rand.rs
forbesmyester: Actually it seems this function is not needed. So deleting.

src/rand.rs
forbesmyester: Well this was/is it's state in `fc_util`, but I agree.

src/rand.rs
forbesmyester: Sure, my reason was that we'd loose the history as Git cannot track the import. You're right it probably won't be maintained so removal is probably the right call.

src/rand.rs
forbesmyester: Well this is a straight port of `firecracker/fc_util.rs` and I don't really understand it... However the function is not required so removed.

src/rand.rs
forbesmyester: `//!`? done

src/rand.rs
forbesmyester: done

src/rand.rs
forbesmyester: Agree, I've added comments saying as such and also changed function names to ensure people understand it's a pseudo random, not the real thing.

forbesmyester:
Aside from the fact it's obvious I wrote the commit message in VIM. I think this addresses all the points and hopefully is acceptable.

src/rand.rs
andreeaflorescu: We do use this function in Firecracker - jailer.

src/rand.rs
andreeaflorescu: I know this was as is in fc_util, but we have different requirements for things that end up in rust-vmm. One example is the mandatory documentation comment for public functions. If you need help with this let me know.

src/rand.rs
andreeaflorescu: git-blaming @aghecenco. Can you helps us with some comments on this function please?

aghecenco: The function was removed from this PR, and also from a PR opened in Firecracker - https://github.com/firecracker-microvm/firecracker/pull/1174 - so there's not much use in documenting it, I think

src/rand.rs
forbesmyester: This is not Firecracker and it's not required for my `tempfile` work.

Thus I don't see a reason to pull it over into this repo.

src/rand.rs
forbesmyester: I agreed and did it.

src/rand.rs
andreeaflorescu: Ok.

src/rand.rs
andreeaflorescu: Yap.

andreeaflorescu:
Changes look good to me. Please squash your commits into one and use the [50/72 rule for formatting commit messages](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule) so we can merge it.

forbesmyester:
rebasing / force-push etc for single commit.

comment is following 50/72 rule etc. Think it's ready to go.

forbesmyester:
Can these tests be re-ran. I'm pretty sure that the code that failed is the same as the code that ran last time and succeeded as `cargo test` does locally. I also don't know how to replicate the `unittests-gnu-arm` failing test, so that does not help.

Assistance would be appreciated.

Thanks

jiangliu:
> Can these tests be re-ran. I'm pretty sure that the code that failed is the same as the code that ran last time and succeeded as `cargo test` does locally. I also don't know how to replicate the `unittests-gnu-arm` failing test, so that does not help.
> 
> Assistance would be appreciated.
> 
> Thanks

Just make minor changes to commit message and force push again.
It's maybe a temporary issue caused by env.

andreeaflorescu:
@jiangliu I can re-trigger the CI without force pushes. But since it failed, maybe it is a good chance to be a bit more descriptive in the commit message.

@forbesmyester I would add something like:
`Adding `rand` module which exports functionality for generating pseudo-random numbers and strings` to the commit message description.

andreeaflorescu:
@forbesmyester we have an issue with the CI right now. I added a PR that fixes the problem (#22). Sorry for the delay!

By the way, the commit message summary should be split on two lines because it has more than 72 chars.

forbesmyester:
> @forbesmyester we have an issue with the CI right now. I added a PR that fixes the problem (#22). Sorry for the delay!
> 
> By the way, the commit message summary should be split on two lines because it has more than 72 chars.

andreeaflorescu:
@forbesmyester we now have a coverage test in our CI. You PR is increasing the coverage and we have a manual process to update it in this case. The file `coverage_config.json` needs to be updated with the new coverage value which is 69.3.

forbesmyester:
Generally, if I change code and stuff breaks, I tend to think it's me... However this code is not called by anything in this project, or anywhere else and it's the same as what has passed before, a good few times.

It is still failing on ARM for reasons that are not apparent. I could set up an ARM based system to test but I don't know where I could find the docs on how to do that... Do I set up Raspbian on an Raspberry Pi or some ARM AWS instance...

It does not seem possible for me to move this forwards so I think you're going to need to take over at least until we're more sure about the ARM stuff.

andreeaflorescu:
We didn't investigate the arm problem. I have some suspicions it might be because of the ARM instance we are using. For some reason we lose connectivity. I am really sorry that the submission process was not smooth. We are still in early stages of the project so not every thing is battle tested. I understand how this can cause frustration for people contributing. If you don't mind, I can push to your branch in case any other changes are required. What do you say?

forbesmyester:
@andreeaflorescu don't worry about it, not everything is perfect all the time. I'm sure you and the team will get it sorted out soon.

It'd be great if you merge it, but I'm not going to say you or anyone else should, that's not my call... 

I'm really just saying I can't solve the CI issue, which I think we're both pretty much sure is the only issue left. Someone else is going to have to take over or get it in or merge it when appropriate.

Thanks for your help getting it this far. Let me know when I can start work on the `tempfile` bit.

jiangliu:
> Generally, if I change code and stuff breaks, I tend to think it's me... However this code is not called by anything in this project, or anywhere else and it's the same as what has passed before, a good few times.
> 
> It is still failing on ARM for reasons that are not apparent. I could set up an ARM based system to test but I don't know where I could find the docs on how to do that... Do I set up Raspbian on an Raspberry Pi or some ARM AWS instance...
> 
> It does not seem possible for me to move this forwards so I think you're going to need to take over at least until we're more sure about the ARM stuff.

Thanks for your patience:)
Seems you have hit a bug in syslog and it should have been fixed by #22 .

forbesmyester:
> Thanks for your patience:)
> Seems you have hit a bug in syslog and it should have been fixed by #22 .

I hope so, but the CI failed with syslog errors when all those commits are already in this branch.

What happens now? Do I re-merge in the master changes and someone else presses the merge button?

jiangliu:
> > Thanks for your patience:)
> > Seems you have hit a bug in syslog and it should have been fixed by #22 .
> 
> I hope so, but the CI failed with syslog errors when all those commits are already in this branch.
> 
> What happens now? Do I re-merge in the master changes and someone else presses the merge button?

Really appreciate if you could help to merge the latest commits from the master. It's a little busy with the vmm-sys-util crates and caused some inconvenience:(

PR #17, opened 2019-07-23
ioctl documentation improvement

liujing2:
@andreeaflorescu @jiangliu @sameo 
Note: In ioctl.rs, there exists a set of definitions copied from kernel but no documentation, which needs more thinking.

### DIFF
--- a/src/ioctl.rs
+++ b/src/ioctl.rs
@@ -8,12 +8,22 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE-BSD-3-Clause file.
 
-//! Macros and wrapper functions for dealing with ioctls.
+//! Macros and functions for working with
+//! [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html).
 use libc;
 use std::os::raw::{c_int, c_uint, c_ulong, c_void};
 use std::os::unix::io::AsRawFd;
 
-/// Raw macro to declare the expression that calculates an ioctl number
+/// Expression that calculates an ioctl number.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+/// # use std::os::raw::c_uint;
+/// use vmm_sys_util::ioctl::_IOC_NONE;
+///
+/// const KVMIO: c_uint = 0xAE;
+/// ioctl_expr!(_IOC_NONE, KVMIO, 0x01, 0);
+/// ```
 #[macro_export]
 macro_rules! ioctl_expr {
     ($dir:expr, $ty:expr, $nr:expr, $size:expr) => {
@@ -24,7 +34,16 @@ macro_rules! ioctl_expr {
     };
 }
 
-/// Raw macro to declare a function that returns an ioctl number.
+/// Declare a function that returns an ioctl number.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+/// # use std::os::raw::c_uint;
+/// use vmm_sys_util::ioctl::_IOC_NONE;
+///
+/// const KVMIO: c_uint = 0xAE;
+/// ioctl_ioc_nr!(KVM_CREATE_VM, _IOC_NONE, KVMIO, 0x01, 0);
+/// ```
 #[macro_export]
 macro_rules! ioctl_ioc_nr {
     ($name:ident, $dir:expr, $ty:expr, $nr:expr, $size:expr) => {
@@ -44,6 +63,13 @@ macro_rules! ioctl_ioc_nr {
 }
 
 /// Declare an ioctl that transfers no data.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+/// # use std::os::raw::c_uint;
+/// const KVMIO: c_uint = 0xAE;
+/// ioctl_io_nr!(KVM_CREATE_VM, KVMIO, 0x01);
+/// ```
 #[macro_export]
 macro_rules! ioctl_io_nr {
     ($name:ident, $ty:expr, $nr:expr) => {
@@ -55,6 +81,12 @@ macro_rules! ioctl_io_nr {
 }
 
 /// Declare an ioctl that reads data.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+/// const TUNTAP: ::std::os::raw::c_uint = 0x54;
+/// ioctl_ior_nr!(TUNGETFEATURES, TUNTAP, 0xcf, ::std::os::raw::c_uint);
+/// ```
 #[macro_export]
 macro_rules! ioctl_ior_nr {
     ($name:ident, $ty:expr, $nr:expr, $size:ty) => {
@@ -79,6 +111,12 @@ macro_rules! ioctl_ior_nr {
 }
 
 /// Declare an ioctl that writes data.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+/// const TUNTAP: ::std::os::raw::c_uint = 0x54;
+/// ioctl_iow_nr!(TUNSETQUEUE, TUNTAP, 0xd9, ::std::os::raw::c_int);
+/// ```
 #[macro_export]
 macro_rules! ioctl_iow_nr {
     ($name:ident, $ty:expr, $nr:expr, $size:ty) => {
@@ -103,6 +141,12 @@ macro_rules! ioctl_iow_nr {
 }
 
 /// Declare an ioctl that reads and writes data.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+/// const VHOST: ::std::os::raw::c_uint = 0xAF;
+/// ioctl_iowr_nr!(VHOST_GET_VRING_BASE, VHOST, 0x12, ::std::os::raw::c_int);
+/// ```
 #[macro_export]
 macro_rules! ioctl_iowr_nr {
     ($name:ident, $ty:expr, $nr:expr, $size:ty) => {
@@ -154,17 +198,107 @@ type IoctlRequest = c_int;
 #[cfg(not(target_env = "musl"))]
 type IoctlRequest = c_ulong;
 
-/// Run an ioctl with no arguments.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with no arguments.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
+/// to call the ioctl.
+/// * `req`: a device-dependent request code.
+///
+/// # Safety
+///
+/// The caller should ensure to pass a valid file descriptor and have the
+/// return value checked.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate libc;
+/// #[macro_use] extern crate vmm_sys_util;
+/// #
+/// # use libc::{open, O_CLOEXEC, O_RDWR};
+/// # use std::fs::File;
+/// # use std::os::raw::{c_char, c_uint};
+/// # use std::os::unix::io::FromRawFd;
+/// use vmm_sys_util::ioctl::ioctl;
+///
+/// const KVMIO: c_uint = 0xAE;
+/// const KVM_API_VERSION: u32 = 12;
+/// ioctl_io_nr!(KVM_GET_API_VERSION, KVMIO, 0x00);
+///
+/// let open_flags = O_RDWR | O_CLOEXEC;
+/// let kvm_fd = unsafe { open("/dev/kvm\0".as_ptr() as *const c_char, open_flags) };
+///
+/// let ret = unsafe { ioctl(&File::from_raw_fd(kvm_fd), KVM_GET_API_VERSION()) };
+///
+/// assert_eq!(ret as u32, KVM_API_VERSION);
+/// ```
 pub unsafe fn ioctl<F: AsRawFd>(fd: &F, req: c_ulong) -> c_int {
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, 0)
 }
 
-/// Run an ioctl with a single value argument.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with a single value argument.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
+/// to call the ioctl.
+/// * `req`: a device-dependent request code.
+/// * `arg`: a single value passed to ioctl.
+///
+/// # Safety
+///
+/// The caller should ensure to pass a valid file descriptor and have the
+/// return value checked.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate libc;
+/// #[macro_use] extern crate vmm_sys_util;
+/// # use libc::{open, O_CLOEXEC, O_RDWR};
+/// # use std::fs::File;
+/// # use std::os::raw::{c_char, c_uint, c_ulong};
+/// # use std::os::unix::io::FromRawFd;
+/// use vmm_sys_util::ioctl::ioctl_with_val;
+///
+/// const KVMIO: c_uint = 0xAE;
+/// const KVM_CAP_USER_MEMORY: u32 = 3;
+/// ioctl_io_nr!(KVM_CHECK_EXTENSION, KVMIO, 0x03);
+///
+/// let open_flags = O_RDWR | O_CLOEXEC;
+/// let kvm_fd = unsafe { open("/dev/kvm\0".as_ptr() as *const c_char, open_flags) };
+///
+/// let ret = unsafe {
+///     ioctl_with_val(
+///         &File::from_raw_fd(kvm_fd),
+///         KVM_CHECK_EXTENSION(),
+///         KVM_CAP_USER_MEMORY as c_ulong,
+///     )
+/// };
+/// assert!(ret > 0);
+/// ```
 pub unsafe fn ioctl_with_val<F: AsRawFd>(fd: &F, req: c_ulong, arg: c_ulong) -> c_int {
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, arg)
 }
 
-/// Run an ioctl with an immutable reference.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with an immutable reference.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
+/// to call the ioctl.
+/// * `req`: a device-dependent request code.
+/// * `arg`: an immutable reference passed to ioctl.
+///
+/// # Safety
+///
+/// The caller should ensure to pass a valid file descriptor and have the
+/// return value checked.
 pub unsafe fn ioctl_with_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &T) -> c_int {
     libc::ioctl(
         fd.as_raw_fd(),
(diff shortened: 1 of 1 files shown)


src/ioctl.rs
```diff
@@ -13,7 +13,19 @@ use libc;
...
-/// Raw macro to declare the expression that calculates an ioctl number
+/// Raw macro to declare the expression that calculates an ioctl number.
+///
+/// ```
+/// # #[macro_use] extern crate vmm_sys_util;
+///
+/// # use std::os::raw::c_uint;
+/// use vmm_sys_util::ioctl::_IOC_NONE;
+///
+/// # fn main() {
```
andreeaflorescu: You don't need `fn main()`. Please remove it. Same for all the other examples.

src/lib.rs
```diff
@@ -6,6 +6,8 @@ extern crate libc;
 #[macro_use]
 pub mod syslog;
 
+/// Macros and functions for working with
+/// [`ioctl`] (http://man7.org/linux/man-pages/man2/ioctl.2.html)
```
andreeaflorescu: There is an extra space between the URL name and link.

Please run `cargo doc` to make sure that the documentation is properly formatted.

src/lib.rs
```diff
@@ -6,6 +6,8 @@ extern crate libc;
 #[macro_use]
 pub mod syslog;
 
+/// Macros and functions for working with
```
andreeaflorescu: If you're adding the documentation for the module here you don't need to also add it to `src/ioctl.rs` because it results in duplicated documentation:
duplicated docs

We should decide if we want the module documentation to be placed in src/lib.rs or in the file where the module is defined.

src/ioctl.rs
```diff
@@ -55,6 +90,15 @@ macro_rules! ioctl_io_nr {
 }
 
 /// Declare an ioctl that reads data.
```
andreeaflorescu: Let's keep the same way of talking about what a macro does across the whole module. The other macro definitions (in vmm-sys-utils) are written as:
```A macro for declaring an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html) that reads data.```

For making it extra neat I would phrase them all the same.

File: src/ioctl.rs [hunk 5]
```diff
@@ -13,7 +13,19 @@ use libc;
 use std::os::raw::{c_int, c_uint, c_ulong, c_void};
 use std::os::unix::io::AsRawFd;
 
-/// Raw macro to declare the expression that calculates an ioctl number
+/// Raw macro to declare the expression that calculates an ioctl number.
```
andreeaflorescu: What is a "raw macro"? Why not use just "macro"?

src/ioctl.rs
```diff
@@ -182,12 +348,68 @@ pub unsafe fn ioctl_with_mut_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &mut
...
+/// let file = unsafe { File::from_raw_fd(0) };
+///
+/// let ret = unsafe { ioctl_with_ptr(&file, TUNSETOFFLOAD, bytes.as_ptr()) };
+/// ```
 pub unsafe fn ioctl_with_ptr<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: *const T) -> c_int {
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, arg as *const c_void)
 }
 
-/// Run an ioctl with a mutable raw pointer.
+/// A wrapper of unsafe [libc::ioctl] with a mutable raw pointer.
```
andreeaflorescu: ```[libc::ioctl]``` is an empty link. I think we should keep the previous documentation: ``` Run an ioctl with a mutable raw pointer.```

src/ioctl.rs
```diff
@@ -182,12 +348,68 @@ pub unsafe fn ioctl_with_mut_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &mut
...
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor.
```
andreeaflorescu: The `fd` is not any opened file descriptor. It is the open file descriptor corresponding to the device on which we want to call the ioctl. Please update the documentation for all the other wrappers over ioctls as well.

src/ioctl.rs
```diff
@@ -154,17 +216,94 @@ type IoctlRequest = c_int;
 #[cfg(not(target_env = "musl"))]
 type IoctlRequest = c_ulong;
 
-/// Run an ioctl with no arguments.
+/// A wrapper of unsafe [libc::ioctl] with no more arguments.
```
andreeaflorescu: "Run an ioctl with no arguments" was better.

src/ioctl.rs
```diff
@@ -154,17 +216,94 @@ type IoctlRequest = c_int;
...
+/// const TUNSETOFFLOAD: c_ulong = 0x4004_54d0;
+/// let file = unsafe { File::from_raw_fd(0) };
+/// let ret = unsafe { ioctl(&file, TUNSETOFFLOAD) };
+/// ```
 pub unsafe fn ioctl<F: AsRawFd>(fd: &F, req: c_ulong) -> c_int {
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, 0)
 }
 
-/// Run an ioctl with a single value argument.
+/// A wrapper of unsafe [libc::ioctl] with a single value argument.
```
andreeaflorescu: "Run an ioctl with a single value argument."

src/ioctl.rs
andreeaflorescu: This stands for all the other comments for the ioctl wrappers.

liujing2: [libc::ioctl] is automatically linked to the right place when we kick it as I saw. 
So we would like not point out what ioctl actually is here?

src/ioctl.rs
liujing2: Got it.

src/ioctl.rs
andreeaflorescu: When I run `cargo doc --open` the `[libc::ioctl]` does not point to anything. Furthermore, it is printed as `[libc::ioctl]`. I think you can just leave it as it was before or add a link to http://man7.org/linux/man-pages/man2/ioctl.2.html

liujing2: Strange... Sure, let's just keep it as before.

liujing2:
@andreeaflorescu Comments are addressed. Thanks.

src/ioctl.rs
andreeaflorescu: @liujing2 have you run `cargo doc --open`?

liujing2: yes, of course. I can kick the link in my side.

src/lib.rs
jiangliu: It would be better to put the module doc into the corresponding source file.

andreeaflorescu: I also prefer that option because for some modules you might have extended documentation and it's making it easier to read the code.

src/ioctl.rs
```diff
@@ -173,7 +298,36 @@ pub unsafe fn ioctl_with_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &T) -> c_
     )
 }
 
-/// Run an ioctl with a mutable reference.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with a mutable reference.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device that
```
andreeaflorescu: How about: "an open file corresponding to the device on which to call the ioctl command"?
Any other suggestions are more than welcomed.

liujing2: That's better. Would you mind to s/ioctl command/ioctl? Not sure if it makes sense. 
I just updated PR with public module descriptor moving into ioctl.rs file.

src/ioctl.rs
```diff
@@ -154,17 +198,100 @@ type IoctlRequest = c_int;
...
+/// ```
+/// extern crate vmm_sys_util;
+/// #
+/// # use std::fs::File;
+/// # use std::os::raw::c_ulong;
+/// # use std::os::unix::io::FromRawFd;
+/// use vmm_sys_util::ioctl::ioctl;
+///
+/// const TUNSETOFFLOAD: c_ulong = 0x4004_54d0;
+/// let file = unsafe { File::from_raw_fd(0) };
```
andreeaflorescu: I just noticed that this example does not actually work. In the documentation, in the safety section you say that the caller should check the return value and that the fd should be valid. 

All the examples on ioctl functions will not work becasue:
- you are using the same ioctl, but that ioctl cannot be called with the parameters you are using
- you are not opening `/dev/net/tun` file

I suggest using kvm ioctls instead as it is easier to get the code working.

liujing2: One question is I found kvm-ioctls crate already has the same file as here. Do we still need re-publish it in this crate? I am thinking maybe other hypervisors would use...?

andreeaflorescu: Once we publish vmm-sys-utils to crates.io I will remove the `ioctl` file from kvm-ioctls and use the ioctl defines from here.

liujing2: Got it. Thanks for explanation.

liujing2: kvm-ioctl are used for examples. Updated.

andreeaflorescu: I didn't really mean to use the `kvm-ioctls` crate. That would result in a cycle dependency as `kvm-ioctls` will eventually use `vmm-sys-utils` for the ioctl macros and functions. I am sorry if I was not explicit enough. I meant using the ioctls on the KVM device. We can use the simplest ioctls we can find that don't need many defines.

I can add a commit on top of your commit because I feel this is wasting too many cycles. What do you say?

liujing2: Of course, I'm fine with that. 
But I think the tests for `ioctl()` and `ioctl_with_val()` are just simple tests for raw ioctl, there's no usage for other vmm-sys-utils in cycle.
And for others, since we want to pass the real and meaningful test parameters, those use `ioctl()` and `ioctl_with_val` in a cycle. Do you feel it does not make sense?

andreeaflorescu: I think is better to have no examples at all rather than examples that are not working. If you think adding examples that work is too complicated/adds too much code, we can make an exception for this module. What do you say?

liujing2: I feel fine for both adding or not. We can keep the ioctl and ioctl_with_val ones because they can give an example for ioctl series functions and are also meaningful,, if you feel other are too complicated.

andreeaflorescu: I personally don't think it is complicated. But you need to do a bit of research to understand which ioctls are appropriate for examples. If you don't know already some ioctls that would be a good match for the functions, I don't think it is worth it to invest time in it.

We can keep `ioctl` and `ioctl_with_val` only if you remove the dependencies on `kvm-bindings` and `kvm-ioctls`. This can be done by:
- use File::open("/dev/kvm") instead of Kvm::new()
- define KVM_CAP_USER_MEMORY in the example, don't use it from kvm-bindings.
- define KVM_GET_API_VERSION in the example, don't use it from kvm-bindings.

src/ioctl.rs
```diff
@@ -154,17 +198,103 @@ type IoctlRequest = c_int;
...
+/// # use std::os::unix::io::FromRawFd;
+/// use vmm_sys_util::ioctl::ioctl_with_val;
+///
+/// let open_flags = O_RDWR | O_CLOEXEC;
+/// let kvm_fd = unsafe { open("/dev/kvm\0".as_ptr() as *const c_char, open_flags) };
+/// const KVMIO: c_uint = 0xAE;
+/// const KVM_CAP_USER_MEMORY: u32 = 3;
+/// ioctl_io_nr!(KVM_CHECK_EXTENSION, KVMIO, 0x03);
+///
+/// let ret = unsafe {
```
andreeaflorescu: Missing assert on ret.

src/ioctl.rs
```diff
@@ -8,12 +8,22 @@
...
 
-//! Macros and wrapper functions for dealing with ioctls.
+//! Macros and functions for working with
+//! [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html).
 use libc;
 use std::os::raw::{c_int, c_uint, c_ulong, c_void};
 use std::os::unix::io::AsRawFd;
 
-/// Raw macro to declare the expression that calculates an ioctl number
+/// Macro to declare the expression that calculates an ioctl number.
```
aghecenco: I'd leave out the `Macro that...` prefix because, in the generated documentation, these are already grouped under the `Macros` category (check `cargo doc --open`).

Generated doc:
```
Macro vmm_sys_util::ioctl_expr
[+] Show declaration
[−]

Macro to declare the expression that calculates an ioctl number.

use vmm_sys_util::ioctl::_IOC_NONE;

const KVMIO: c_uint = 0xAE;
ioctl_expr!(_IOC_NONE, KVMIO, 0x01, 0);
```

```suggestion
/// Expression that calculates an ioctl number.
```

src/ioctl.rs
```diff
@@ -24,7 +34,16 @@ macro_rules! ioctl_expr {
     };
 }
 
-/// Raw macro to declare a function that returns an ioctl number.
+/// Macro to declare a function that returns an ioctl number.
```
aghecenco:
```suggestion
/// Declare a function that returns an ioctl number.
```

src/ioctl.rs
```diff
@@ -43,7 +62,14 @@ macro_rules! ioctl_ioc_nr {
     };
 }
 
-/// Declare an ioctl that transfers no data.
+/// Macro to declare an ioctl that transfers no data.
```
aghecenco:
```suggestion
/// Declare an ioctl that transfers no data.
```

src/ioctl.rs
```diff
@@ -54,7 +80,13 @@ macro_rules! ioctl_io_nr {
     };
 }
 
-/// Declare an ioctl that reads data.
+/// Macro to declare an ioctl that reads data.
```
aghecenco:
```suggestion
/// Declare an ioctl that reads data.
```

src/ioctl.rs
```diff
@@ -78,7 +110,13 @@ macro_rules! ioctl_ior_nr {
     };
 }
 
-/// Declare an ioctl that writes data.
+/// Macro to declare an ioctl that writes data.
```
aghecenco:
```suggestion
/// Declare an ioctl that writes data.
```

src/ioctl.rs
```diff
@@ -102,7 +140,13 @@ macro_rules! ioctl_iow_nr {
     };
 }
 
-/// Declare an ioctl that reads and writes data.
+/// Macro to declare an ioctl that reads and writes data.
```
aghecenco:
```suggestion
/// Declare an ioctl that reads and writes data.
```

src/ioctl.rs
```diff
@@ -154,17 +198,104 @@ type IoctlRequest = c_int;
 #[cfg(not(target_env = "musl"))]
 type IoctlRequest = c_ulong;
 
-/// Run an ioctl with no arguments.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with no arguments.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
```
aghecenco: I find this description a bit too verbose, `ioctl` users expect the fd to be the "destination" for the ioctl.
```suggestion
/// * `fd`: open device fd.
```

src/ioctl.rs
```diff
@@ -154,17 +198,104 @@ type IoctlRequest = c_int;
...
+/// #
+/// # use libc::{open, O_CLOEXEC, O_RDWR};
+/// # use std::fs::File;
+/// # use std::os::raw::{c_char, c_uint};
+/// # use std::os::unix::io::FromRawFd;
+/// use vmm_sys_util::ioctl::ioctl;
+///
+/// let open_flags = O_RDWR | O_CLOEXEC;
+/// let kvm_fd = unsafe { open("/dev/kvm\0".as_ptr() as *const c_char, open_flags) };
+/// const KVMIO: c_uint = 0xAE;
```
aghecenco: nit: move the `const` declarations separately from the fd/ioctl logic, to neatly delimit the functional blocks 
```rust
#[macro_use] extern crate vmm_sys_util;
use vmm_sys_util::ioctl::ioctl;

const KVMIO: c_uint = 0xAE;
const KVM_API_VERSION: u32 = 12;
ioctl_io_nr!(KVM_GET_API_VERSION, KVMIO, 0x00);

let open_flags = O_RDWR | O_CLOEXEC;
let kvm_fd = unsafe { open("/dev/kvm\0".as_ptr() as *const c_char, open_flags) };

let ret = unsafe { ioctl(&File::from_raw_fd(kvm_fd), KVM_GET_API_VERSION()) };

assert_eq!(ret as u32, KVM_API_VERSION);
```

This applies to all the ioctl examples, I won't post individual nit comments with code suggestions.

src/ioctl.rs
```diff
@@ -154,17 +198,104 @@ type IoctlRequest = c_int;
...
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, 0)
 }
 
-/// Run an ioctl with a single value argument.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with a single value argument.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
```
aghecenco:
```suggestion
/// * `fd`: open device fd.
```

src/ioctl.rs
```diff
@@ -154,17 +198,104 @@ type IoctlRequest = c_int;
...
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, arg)
 }
 
-/// Run an ioctl with an immutable reference.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with an immutable reference.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
```
aghecenco:
```suggestion
/// * `fd`: open device fd.
```

src/ioctl.rs
```diff
@@ -173,7 +304,20 @@ pub unsafe fn ioctl_with_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &T) -> c_
     )
 }
 
-/// Run an ioctl with a mutable reference.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with a mutable reference.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
```
aghecenco:
```suggestion
/// * `fd`: open device fd.
```

src/ioctl.rs
```diff
@@ -182,12 +326,38 @@ pub unsafe fn ioctl_with_mut_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &mut
     )
 }
 
-/// Run an ioctl with a raw pointer.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with a raw pointer.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
```
aghecenco:
```suggestion
/// * `fd`: open device fd.
```

src/ioctl.rs
```diff
@@ -182,12 +326,38 @@ pub unsafe fn ioctl_with_mut_ref<F: AsRawFd, T>(fd: &F, req: c_ulong, arg: &mut
...
     libc::ioctl(fd.as_raw_fd(), req as IoctlRequest, arg as *const c_void)
 }
 
-/// Run an ioctl with a mutable raw pointer.
+/// Run an [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html)
+/// with a mutable raw pointer.
+///
+/// # Arguments
+///
+/// * `fd`: an open file descriptor corresponding to the device on which
```
aghecenco:
```suggestion
/// * `fd`: open device fd.
```

src/ioctl.rs
andreeaflorescu: I would say to leave this one as it is because it is closer to the [ioctl documentation](http://man7.org/linux/man-pages/man2/ioctl.2.html). In this case I don't think more details hurt.

src/ioctl.rs
liujing2: Thanks for your reviewing.  I'll update soon.

src/ioctl.rs
liujing2: Comments are addressed, thanks!

src/ioctl.rs
liujing2: I've amend the changes into commit. Thanks!

PR #18, opened 2019-07-23
eventfd improvements on documentation

liujing2:
Extend detailed documentation for eventfd module. 
Add examples for each public interface.

### DIFF
--- a/src/eventfd.rs
+++ b/src/eventfd.rs
@@ -5,21 +5,36 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE-BSD-3-Clause file.
 
+//! Structure and wrapper functions for working with
+//! [`eventfd`](http://man7.org/linux/man-pages/man2/eventfd.2.html).
+
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


liujing2:
@jiangliu  I moved the public module descriptor from lib.rs into eventfd.rs file. Could you review again this little change? Thanks!

liujing2:
Comments are addressed.

src/eventfd.rs
```diff
@@ -5,21 +5,36 @@
...
-/// A safe wrapper around a Linux eventfd (man 2 eventfd).
+/// A safe wrapper around Linux
+/// [`eventfd`](http://man7.org/linux/man-pages/man2/eventfd.2.html).
 pub struct EventFd {
     eventfd: File,
 }
 
 impl EventFd {
-    /// Creates a new blocking EventFd with an initial value.
+    /// Create a new blocking EventFd with an initial value.
```
andreeaflorescu: The eventfd is not necessarily blocking as you can pass flags in which you can specify `EFD_NONBLOCK`. We can just remove the word "blocking" for the documentation to be accurate.

PR #19, opened 2019-07-24
license: improvements on license

liujing2:
The crate needs Apache-2.0 or BSD-3-Clause but some files are not correctly
described with the license. Make all files correct on license declaration.

### DIFF
--- a/src/eventfd.rs
+++ b/src/eventfd.rs
@@ -1,9 +1,8 @@
 // Copyright 2019 Intel Corporation. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
 //
 // Copyright 2017 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
 
 //! Structure and wrapper functions for working with
 //! [`eventfd`](http://man7.org/linux/man-pages/man2/eventfd.2.html).
--- a/src/fallocate.rs
+++ b/src/fallocate.rs
@@ -1,3 +1,9 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+//
+// SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
+
 //! Enum and function for dealing with an allocated disk space
 //! by [`fallocate`](http://man7.org/linux/man-pages/man2/fallocate.2.html).
 
--- a/src/file_traits.rs
+++ b/src/file_traits.rs
@@ -1,8 +1,8 @@
-// Copyright 2018 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
+// Copyright 2019 Intel Corporation. All Rights Reserved.
 //
-// SPDX-License-Identifier: BSD-3-Clause
+// Copyright 2018 The Chromium OS Authors. All rights reserved.
+//
+// SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
 
 //! Traits for handling file synchronization and length.
 
--- a/src/seek_hole.rs
+++ b/src/seek_hole.rs
@@ -1,8 +1,8 @@
-// Copyright 2018 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
+// Copyright 2019 Intel Corporation. All Rights Reserved.
 //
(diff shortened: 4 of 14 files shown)


src/signal.rs
```diff
@@ -1,5 +1,5 @@
 // Copyright 2019 Intel Corporation. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 //
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
```
jiangliu: This line should be removed too.

liujing2: Line 5 is copied from Amazon license together with line 4. Do you mean we only tell the companies, removing the original license that companies used, and leave one line for declaring final license?

jiangliu: not sure whether it makes sense to include multiple "SPDX-License-Identifier" header:)

liujing2: @andreeaflorescu @sameo Do you know the good way to list the companies and licenses that might be different ?

andreeaflorescu: @liujing2 I really don't know what is the best approach here. I think it makes sense to send an email to the rust-vmm email list and ask there.

sameo: The goal of the SPDX id is to provide a single, simple and unified way to expose under which terms a source code file is licensed. Havinf multiple `SPDX-License-Identifier` is confusing and goes against the point of using SPDX ids. So I'd go with a single one.

In that case, Amazon licensed their piece of code under the Apache 2.0 license (exclusively, not dual licensed). So unless Amazon wants to re license (for dual licensing), I think the single SPDX identifier that we should use here is:

```
SPDX-License-Identifier: Apache-2.0
```

sameo: If the piece of code also includes some BSD-3-Clause code from crosvm, then we should use:

```
SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
```

liujing2: Thanks for your guidance.  The PR is updated. :)

src/eventfd.rs
```diff
@@ -1,9 +1,8 @@
 // Copyright 2019 Intel Corporation. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
 //
 // Copyright 2017 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: BSD-3-Clause
```
sameo: This is APACHE AND BSD, as I assumed Intel added Apache licensed code.

src/fallocate.rs
```diff
@@ -1,3 +1,9 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+//
+// SPDX-License-Identifier: BSD-3-Clause
```
sameo: Ditto.

src/file_traits.rs
```diff
@@ -1,6 +1,6 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
 // Copyright 2018 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
 //
 // SPDX-License-Identifier: BSD-3-Clause
```
sameo: Ditto.

src/seek_hole.rs
```diff
@@ -1,6 +1,6 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
 // Copyright 2018 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
 //
 // SPDX-License-Identifier: BSD-3-Clause
```
sameo: Ditto.

src/syslog.rs
```diff
@@ -1,9 +1,8 @@
 // Copyright 2019 Intel Corporation. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
 //
 // Copyright 2017 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: BSD-3-Clause
```
sameo: APACHE AND BSD

src/tempdir.rs
```diff
@@ -1,6 +1,6 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
+//
 // Copyright 2017 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
 //
 // SPDX-License-Identifier: BSD-3-Clause
```
sameo: Ditto.

Issue #13, opened 2019-07-17
Update Crate Metadata


andreeaflorescu:
There are a few things missing from `Cargo.toml`. Some of them are required for publishing (like description), while others are nice to have.
- description
- repository
- readme
- keywords (this is making the package easy to find on crates.io)

For reference, you can check the [Publishing a Crate to Crates.io](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#adding-metadata-to-a-new-crate).

PR #29, opened 2019-08-01
Add metadata and codeowners 

liujing2:
Please point to me if anything not correct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,12 @@
 [package]
 name = "vmm-sys-util"
 version = "0.1.0"
-authors = ["Jing Liu <jing2.liu@linux.intel.com>"]
-license = "Apache-2.0"
+authors = ["Intel Virtualization Team <vmm-maintainers@intel.com>"]
+description = "A system utility set"
+repository = "https://github.com/rust-vmm/vmm-sys-util"
+readme = "README.md"
+keywords = ["utils"]
+license = "Apache-2.0 OR BSD-3-Clause"
 
 [dependencies]
 libc = ">=0.2.39"
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @liujing2 @sameo @andreeaflorescu @jiangliu


Cargo.toml
```diff
@@ -1,8 +1,12 @@
 [package]
 name = "vmm-sys-util"
 version = "0.1.0"
-authors = ["Jing Liu <jing2.liu@linux.intel.com>"]
-license = "Apache-2.0"
+authors = ["Intel Virtualization Team <vmm-maintainers@intel.com>"]
+description = "A system utility set for virtualization"
+repository = "https://github.com/rust-vmm/vmm-sys-util"
+readme = "README.md"
+keywords = ["vmm"]
```
andreeaflorescu: I don't think that vmm is the best keyword here because this is more of a "utils" crate and it's not directly related to vmms.

Cargo.toml
```diff
@@ -1,8 +1,12 @@
 [package]
 name = "vmm-sys-util"
 version = "0.1.0"
-authors = ["Jing Liu <jing2.liu@linux.intel.com>"]
-license = "Apache-2.0"
+authors = ["Intel Virtualization Team <vmm-maintainers@intel.com>"]
+description = "A system utility set for virtualization"
```
andreeaflorescu: This is not necessarily for virtualization. I think we should find a better description, but I can't think of any right now.

Cargo.toml
```diff
@@ -1,8 +1,12 @@
 [package]
 name = "vmm-sys-util"
 version = "0.1.0"
-authors = ["Jing Liu <jing2.liu@linux.intel.com>"]
-license = "Apache-2.0"
+authors = ["Intel Virtualization Team <vmm-maintainers@intel.com>"]
```
andreeaflorescu: Please state in the commit message that this is fixing https://github.com/rust-vmm/vmm-sys-util/issues/13

andreeaflorescu: Also for the next commit, specify that it fixes https://github.com/rust-vmm/vmm-sys-util/issues/2

Cargo.toml
liujing2: Yep, it can be used by other scenarios either. What I am thinking is the name vmm-sys-util which describes the reason we create this crate?

liujing2: Maybe let's just make it as "A system utility set". The description is not web-view though.

Cargo.toml
liujing2: Comments are addressed. Thanks!

PR #30, opened 2019-08-01
timerfd: Add unit tests

### LINKED ISSUE #25: timerfd.rs - add unit tests
The coverage as of 30 July 2019 is 67.9%.

The following functions and trait implementations are not tested:
- `clear()`
- `FromRawFd`
- `IntoRawFd`

Also, error cases are also not tested, but I am not sure how feasible it is to test them.

liujing2:
clear(), FromRawFd and IntoRawFd are tested.

Fix #25.

Signed-off-by: Jing Liu <jing2.liu@linux.intel.com>

### DIFF
--- a/src/timerfd.rs
+++ b/src/timerfd.rs
@@ -219,6 +219,19 @@ mod tests {
     use std::time::{Duration, Instant};
 
     #[test]
+    fn test_from_raw_fd() {
+        let ret = unsafe { timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC) };
+        let tfd = unsafe { TimerFd::from_raw_fd(ret) };
+        assert_eq!(tfd.is_armed().unwrap(), false);
+    }
+
+    #[test]
+    fn test_into_raw_fd() {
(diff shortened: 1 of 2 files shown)


jiangliu:
Need to update coverage_config.json to fix the failure:
AssertionError: Coverage is increased from 67.9 to 69.6. Please update the coverage in tests/coverage.

liujing2:
> Need to update coverage_config.json to fix the failure:
> AssertionError: Coverage is increased from 67.9 to 69.6. Please update the coverage in tests/coverage.

Thanks for pointing out. Updated now.

src/timerfd.rs
```diff
@@ -233,6 +246,7 @@ mod tests {
 
         assert_eq!(count, 1);
         assert!(now.elapsed() >= dur);
+        tfd.clear().expect("unable to clear the timer");
```
andreeaflorescu: I don't think that calling the clear function in this case is enough. You should also check that the `clear` function had the expected effect.

liujing2: Fixed.

PR #31, opened 2019-08-07
signal: Add documentation and unit test

### LINKED ISSUE #24: signal.rs - add unit tests
The signal module has a coverage of 31% (as of 31 July 2019).
Most of the functionality is untested such as:
- Display implementation (this should be optional)
- `create_sigset`
- `get_blocked_signals`
- `block_signal`
- `unblock_signal`
- `clear_signal`

Testing signals might be tricky in rust in the unit tests as all unit tests run in the same process. There are some mechanisms by which this problem can be avoided such as spawning a new process for tests that might kill the main process which runs the unit tests.

liujing2:
This fix #24

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -137,6 +137,15 @@ fn SIGRTMAX() -> c_int {
 ///
 /// * `num`: the signal number to be verify.
 /// * `for_vcpu`: if `num` is a vcpu signal.
+///
+/// # Examples
+///
+/// ```
+/// extern crate vmm_sys_util;
+/// use vmm_sys_util::signal::validate_signal_num;
+///
+/// let num = validate_signal_num(1, true).unwrap();
+/// ```
 pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
     if for_vcpu {
         let actual_num = num + SIGRTMIN();
@@ -164,6 +173,21 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
 /// * `for_vcpu`: whether `num` is a vcpu signal.
 /// * `flag`: specify the behavior of the signal.
 ///    Set SA_SIGINFO or SA_RESTART if wants to restart after signal received.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate libc;
+/// extern crate vmm_sys_util;
+/// # use libc::{c_int, c_void, siginfo_t, SA_SIGINFO};
+/// use vmm_sys_util::signal::{register_signal_handler, SignalHandler};
+///
+/// extern "C" fn handle_signal(_: c_int, _: *mut siginfo_t, _: *mut c_void) {}
+/// unsafe {
+///     register_signal_handler(0, SignalHandler::Siginfo(handle_signal), true, SA_SIGINFO);
+/// }
+/// ```
+
 pub unsafe fn register_signal_handler(
     num: c_int,
     handler: SignalHandler,
@@ -188,6 +212,19 @@ pub unsafe fn register_signal_handler(
 /// # Arguments
 ///
 /// * `signals`: signal numbers to be added to the new `sigset`.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate libc;
+/// extern crate vmm_sys_util;
+/// # use libc::sigismember;
+/// use vmm_sys_util::signal::create_sigset;
+///
+/// let sigset = create_sigset(&[1]).unwrap();
+///
+/// unsafe { assert_eq!(sigismember(&sigset, 1), 1); }
+/// ```
 pub fn create_sigset(signals: &[c_int]) -> errno::Result<sigset_t> {
     // sigset will actually be initialized by sigemptyset below.
     let mut sigset: sigset_t = unsafe { mem::zeroed() };
@@ -214,6 +251,16 @@ pub fn create_sigset(signals: &[c_int]) -> errno::Result<sigset_t> {
 /// Use [`pthread_sigmask`](http://man7.org/linux/man-pages/man3/pthread_sigmask.3.html)
 /// to fetch the signal mask which is blocked for the caller, return the signal mask as
 /// a vector of c_int.
+///
+/// # Examples
+///
+/// ```
+/// extern crate vmm_sys_util;
+/// use vmm_sys_util::signal::{block_signal, get_blocked_signals};
+///
+/// block_signal(1).unwrap();
+/// assert!(get_blocked_signals().unwrap().contains(&(1)));
+/// ```
 pub fn get_blocked_signals() -> SignalResult<Vec<c_int>> {
     let mut mask = Vec::new();
 
@@ -244,6 +291,15 @@ pub fn get_blocked_signals() -> SignalResult<Vec<c_int>> {
 /// # Arguments
 ///
 /// * `num`: the signal to be masked.
+///
+/// # Examples
+///
+/// ```
+/// extern crate vmm_sys_util;
+/// use vmm_sys_util::signal::block_signal;
+///
+/// block_signal(1).unwrap();
+/// ```
 pub fn block_signal(num: c_int) -> SignalResult<()> {
     let sigset = create_sigset(&[num]).map_err(Error::CreateSigset)?;
 
@@ -270,6 +326,17 @@ pub fn block_signal(num: c_int) -> SignalResult<()> {
 /// # Arguments
 ///
 /// * `num`: the signal to be unmasked.
+///
+/// # Examples
+///
+/// ```
+/// extern crate vmm_sys_util;
+/// use vmm_sys_util::signal::{block_signal, get_blocked_signals, unblock_signal};
+///
+/// block_signal(1).unwrap();
+/// assert!(get_blocked_signals().unwrap().contains(&(1)));
+/// unblock_signal(1).unwrap();
+/// ```
 pub fn unblock_signal(num: c_int) -> SignalResult<()> {
     let sigset = create_sigset(&[num]).map_err(Error::CreateSigset)?;
 
@@ -286,6 +353,30 @@ pub fn unblock_signal(num: c_int) -> SignalResult<()> {
 /// # Arguments
 ///
 /// * `num`: the signal to be cleared.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate libc;
+/// extern crate vmm_sys_util;
+/// # use libc::{pthread_kill, sigismember, sigpending, sigset_t};
+/// # use std::mem;
+/// # use std::thread;
+/// # use std::time::Duration;
+/// use vmm_sys_util::signal::{block_signal, clear_signal, Killable};
+///
+/// block_signal(1).unwrap();
+/// let killable = thread::spawn(move || {
+///     thread::sleep(Duration::from_millis(100));
+///     unsafe {
+///         let mut chkset: sigset_t = mem::zeroed();
+///         sigpending(&mut chkset);
+///         assert_eq!(sigismember(&chkset, 1), 1);
+///     }
+/// });
+/// unsafe { pthread_kill(killable.pthread_handle(), 1); }
+/// clear_signal(1).unwrap();
+/// ```
 pub fn clear_signal(num: c_int) -> SignalResult<()> {
     let sigset = create_sigset(&[num]).map_err(Error::CreateSigset)?;
 
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -382,11 +382,21 @@ mod tests {
     use std::thread;
     use std::time::Duration;
 
+    // Reserve for each vcpu signal.
     static mut SIGNAL_HANDLER_CALLED: bool = false;
 
     extern "C" fn handle_signal(_: c_int, _: *mut siginfo_t, _: *mut c_void) {
         unsafe {
+            // In the tests, there only uses vcpu signal.
             SIGNAL_HANDLER_CALLED = true;
+        }
+    }
+
+    fn is_pending(signal: c_int) -> bool {
+        unsafe {
+            let mut chkset: sigset_t = mem::zeroed();
+            sigpending(&mut chkset);
+            sigismember(&chkset, signal) == 1
         }
     }
 
@@ -473,5 +483,49 @@ mod tests {
         // previous signal is effectively ignored. If we were to join killable here, we would block
         // forever as the loop keeps running. Since we don't join, the thread will become detached
         // as the handle is dropped, and will be killed when the process/main thread exits.
+    }
+
+    #[test]
+    fn test_block_unblock_signal() {
+        let signal = SIGRTMIN() + 0;
+
+        // Check if it is blocked.
+        unsafe {
+            let mut sigset: sigset_t = mem::zeroed();
+            pthread_sigmask(SIG_BLOCK, null(), &mut sigset as *mut sigset_t);
+            assert_eq!(sigismember(&sigset, signal), 0);
+        }
+
+        block_signal(signal).unwrap();
+        assert!(get_blocked_signals().unwrap().contains(&(signal)));
+
+        unblock_signal(signal).unwrap();
+        assert!(!get_blocked_signals().unwrap().contains(&(signal)));
+    }
+
+    #[test]
+    fn test_clear_pending() {
+        let signal = SIGRTMIN() + 1;
+
+        block_signal(signal).unwrap();
+
+        // Block the signal, which means it won't be delivered until it is
+        // unblocked. Pending between the time when the signal which is set as blocked
+        // is generated and when is delivered.
+        let killable = thread::spawn(move || {
+            loop {
+                // Wait for the signal being killed.
+                thread::sleep(Duration::from_millis(100));
+                if is_pending(signal) {
+                    clear_signal(signal).unwrap();
+                    assert!(!is_pending(signal));
+                    break;
+                }
+            }
+        });
+
+        // Send a signal to the thread.
+        assert!(killable.kill(1).is_ok());
+        killable.join().unwrap();
     }
 }
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -6,6 +6,9 @@
 //
 // SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
 
+//! Enums, traits and functions for working with
+//! [`signal`](http://man7.org/linux/man-pages/man7/signal.7.html).
+
 use libc::{
     c_int, c_void, pthread_kill, pthread_sigmask, pthread_t, sigaction, sigaddset, sigemptyset,
     siginfo_t, sigismember, sigpending, sigset_t, sigtimedwait, timespec, EAGAIN, EINTR, EINVAL,
@@ -21,6 +24,7 @@ use std::ptr::{null, null_mut};
 use std::result;
 use std::thread::JoinHandle;
 
+/// The error cases enumeration for signal handling.
 #[derive(Debug)]
 pub enum Error {
     /// Couldn't create a sigset.
@@ -73,12 +77,17 @@ impl Display for Error {
     }
 }
 
+/// A simplified [Result](https://doc.rust-lang.org/std/result/enum.Result.html) type
+/// for operations that can return [`Error`](Enum.error.html).
 pub type SignalResult<T> = result::Result<T, Error>;
 type SiginfoHandler = extern "C" fn(num: c_int, info: *mut siginfo_t, _unused: *mut c_void) -> ();
 
+/// Enum of signal handlers for
+/// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
 pub enum SignalHandler {
+    /// Corresponds to the `sa_sigaction` member of
+    /// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
     Siginfo(SiginfoHandler),
-    // TODO add a`SimpleHandler` when `libc` adds `sa_handler` support to `sigaction`.
 }
 
 impl SignalHandler {
@@ -86,8 +95,9 @@ impl SignalHandler {
         act.sa_flags = flag;
     }
 }
-/// Fills a `sigaction` structure from of the signal handler.
-/// Refer to http://man7.org/linux/man-pages/man7/signal.7.html
+
+/// Convert a [`SignalHandler`](enum.SignalHandler.html) into a
+/// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html)
 impl Into<sigaction> for SignalHandler {
     fn into(self) -> sigaction {
         let mut act: sigaction = unsafe { mem::zeroed() };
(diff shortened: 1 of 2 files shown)


liujing2:
syslog breaks some CI test...

src/signal.rs
```diff
@@ -75,10 +79,18 @@ impl Display for Error {
...
+///
+/// This typedef is generally used to avoid writing out `Error` directly and is
+/// otherwise a direct mapping to `Result`.
 pub type SignalResult<T> = result::Result<T, Error>;
 type SiginfoHandler = extern "C" fn(num: c_int, info: *mut siginfo_t, _unused: *mut c_void) -> ();
 
+/// Enum of signal handlers for
+/// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
 pub enum SignalHandler {
+    /// Specify the signal handler `sa_sigaction` of sigaction structure.
```
aghecenco:
```suggestion
    /// Corresponds to the `sa_sigaction` member of [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
```

src/signal.rs
```diff
@@ -75,10 +79,18 @@ impl Display for Error {
...
     Siginfo(SiginfoHandler),
     // TODO add a`SimpleHandler` when `libc` adds `sa_handler` support to `sigaction`.
```
aghecenco: I don't know if this is even relevant anymore since the recommended way to handle signals is `sigaction` not `signal`/ I'd remove the comment. And maybe consider removing the `enum` altogether since there'd only be a single variant left.

src/signal.rs
```diff
@@ -88,8 +100,9 @@ impl SignalHandler {
         act.sa_flags = flag;
     }
 }
-/// Fills a `sigaction` structure from of the signal handler.
-/// Refer to http://man7.org/linux/man-pages/man7/signal.7.html
+
+/// Fill a [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html)
```
aghecenco:
```suggestion
/// Convert a [`SignalHandler`](enum.SignalHandler.html) into a 
/// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
```

src/signal.rs
```diff
@@ -107,21 +120,23 @@ extern "C" {
...
 fn SIGRTMAX() -> c_int {
     unsafe { __libc_current_sigrtmax() }
 }
 
-/// Verifies that a signal number is valid: for VCPU signals, it needs to be enclosed within the OS
-/// limits for realtime signals, and the remaining ones need to be between the minimum (SIGHUP) and
-/// maximum (SIGSYS) values.
+/// Verify that a signal number is valid.
+///
+/// For VCPU signal, it needs to be enclosed within the OS
```
aghecenco:
```suggestion
VCPU signals must be enclosed within the OS
```

src/signal.rs
```diff
@@ -134,13 +149,47 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
     Err(errno::Error::new(EINVAL))
 }
 
-/// Registers `handler` as the signal handler of signum `num`.
+/// Register the signal handler of signum.
```
aghecenco:
```suggestion
/// Register the signal handler of `signum`.
```

src/signal.rs
```diff
@@ -134,13 +149,47 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
...
+/// use vmm_sys_util::signal::{register_signal_handler, SignalHandler};
+///
+/// static mut SIGNAL_HANDLER_CALLED: bool = false;
+/// extern "C" fn handle_signal(_: c_int, _: *mut siginfo_t, _: *mut c_void) {
+///     unsafe {
+///         SIGNAL_HANDLER_CALLED = true;
+///     }
+/// }
+///
+/// unsafe {
```
aghecenco: I don't find this example very relevant because it registers itself for signal `0` and never gets to handle it.

src/signal.rs
```diff
@@ -134,13 +149,47 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
...
+/// * `handler`: the signal handler function to register.
+/// * `for_vcpu`: whether `num` is a vcpu signal.
+/// * `flag`: specify the behavior of the signal.
+///    Set SA_SIGINFO or SA_RESTART if wants to restart after signal received.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate libc;
+/// extern crate vmm_sys_util;
```
aghecenco: Please comment everything that isn't relevant to the generated example - like `extern crate` definitions, `use`s etc.

src/signal.rs
```diff
@@ -180,7 +235,11 @@ pub fn create_sigset(signals: &[c_int]) -> errno::Result<sigset_t> {
     Ok(sigset)
 }
 
-/// Retrieves the signal mask of the current thread as a vector of c_ints.
+/// Retrieve the signal mask that is blocked of the current thread.
+///
+/// Use [`pthread_sigmask`](http://man7.org/linux/man-pages/man3/pthread_sigmask.3.html)
+/// to fetch the signal mask which is blocked for the caller, return the signal mask as
+/// a vector of c_int.
```
aghecenco:
```suggestion
/// a vector of `i32`.
```

Stressing out again my previous point of using either `c_int` or `i32` across the code.

src/signal.rs
```diff
@@ -202,10 +261,14 @@ pub fn get_blocked_signals() -> SignalResult<Vec<c_int>> {
     Ok(mask)
 }
 
-/// Masks given signal.
+/// Mask a given signal.
+///
+/// Set the given signal `num` as blocked.
+/// If signal is already blocked, the call will fail with Error::SignalAlreadyBlocked.
```
aghecenco:
```suggestion
/// If signal is already blocked, the call will fail with [`SignalAlreadyBlocked`](enum.Error.html#variant.SignalAlreadyBlocked).
```

src/signal.rs
```diff
@@ -287,22 +359,29 @@ pub fn clear_signal(num: c_int) -> SignalResult<()> {
...
-/// guaranteed to not be used by the C runtime.
+/// Note that this is only useful for signals between SIGRTMIN and SIGRTMAX
+/// because these are guaranteed to not be used by the C runtime.
 ///
-/// This is marked unsafe because the implementation of this trait must guarantee that the returned
-/// pthread_t is valid and has a lifetime at least that of the trait object.
+/// # Safety
+///
+/// This is marked unsafe because the implementation of this trait must
+/// guarantee that the returned pthread_t is valid and has a lifetime at
```
aghecenco:
```suggestion
/// guarantee that the returned `pthread_t` is valid and has a lifetime at
```

src/signal.rs
```diff
@@ -287,22 +359,29 @@ pub fn clear_signal(num: c_int) -> SignalResult<()> {
...
     fn pthread_handle(&self) -> pthread_t;
 
-    /// Sends the signal `num + SIGRTMIN` to this killable thread.
+    /// Send a signal to this killable thread.
+    ///
+    /// # Arguments
     ///
-    /// The value of `num + SIGRTMIN` must not exceed `SIGRTMAX`.
+    /// * `num`: specify the signal and `num + SIGRTMIN` will be sent.
+    /// Note the value of `num + SIGRTMIN` must not exceed SIGRTMAX.
```
aghecenco:
```suggestion
    /// Note the value of `num + SIGRTMIN` must not exceed `SIGRTMAX`.
```

src/signal.rs
```diff
@@ -107,21 +120,23 @@ extern "C" {
...
 }
 
-/// Verifies that a signal number is valid: for VCPU signals, it needs to be enclosed within the OS
-/// limits for realtime signals, and the remaining ones need to be between the minimum (SIGHUP) and
-/// maximum (SIGSYS) values.
+/// Verify that a signal number is valid.
+///
+/// For VCPU signal, it needs to be enclosed within the OS
+/// limits for realtime signals. The remaining ones need
+/// to be between the minimum (SIGHUP) and maximum (SIGSYS) values.
```
aghecenco: Please add an `# Arguments` section.

aghecenco:
```suggestion
/// to be between the minimum (`SIGHUP`) and maximum (`SIGSYS`) values.
```

src/signal.rs
liujing2: I think the enum means there reserves a space for `sa_handler` of `sigaction` structure.

src/signal.rs
liujing2: Fixed

src/signal.rs
liujing2: Fixed.

src/signal.rs
```diff
@@ -75,10 +79,18 @@ impl Display for Error {
     }
 }
 
+/// A specialized [Result](https://doc.rust-lang.org/std/result/enum.Result.html) type
+/// for operations that can return [`Error`]: Enum.error.html.
```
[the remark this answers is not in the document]
liujing2: Fixed.

src/signal.rs
```diff
@@ -8,6 +8,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE-BSD-3-Clause file.
 
+//! Enum, trait and functions for working with
```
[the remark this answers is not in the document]
liujing2: Fixed

src/signal.rs
liujing2: I think enum means reserving a space for `sa_handler` of `sigaction` structure in future.

src/signal.rs
liujing2: Fixed

src/signal.rs
liujing2: Umm, I think it would be good to move the example to block_signal because we can directly show examples for multiple APIs there.

src/signal.rs
liujing2: Right. Fixed.

src/signal.rs
liujing2: This is great. Fixed.

src/signal.rs
liujing2: Fixed

src/signal.rs
liujing2: Fixed

src/signal.rs
```diff
@@ -408,13 +408,19 @@ mod tests {
     use std::time::Duration;
 
     static mut SIGNAL_HANDLER_CALLED: bool = false;
+    static mut SIGNAL_HANDLER_CALLED_FLAG: bool = false;
```
alexandruag: Having these two global variables with similar names (`SIGNAL_HANDLER_CALLED` and `SIGNAL_HANDLER_CALLED_FLAG`), and also the two functions with similar names (`handle_signal_func` and `handle_signal`) is a bit confusing, and might not scale well when we keep adding tests.

We can replace the multiple global variables with an array of `bool`s (or `Atomic<Something>` if we want to get rid of the unsafe), basically one for each signal number. Then we can keep a single handler function which knows the index corresponding to the current signal by inspecting the first parameter it receives (I'm assuming it contains the signal number).

src/signal.rs
liujing2: Sure. Added in update.

liujing2: Fixed

src/signal.rs
```diff
@@ -499,4 +505,89 @@ mod tests {
...
+        let mut sigset: sigset_t = unsafe { mem::zeroed() };
+
+        // Block a signal.
+        unsafe {
+            sigemptyset(&mut sigset);
+            sigaddset(&mut sigset, SIGRTMIN() + 0);
+            pthread_sigmask(SIG_BLOCK, &sigset, null_mut());
+        }
+
+        assert!(get_blocked_signals().unwrap().contains(&(SIGRTMIN() + 0)));
```
alexandruag: Can we test `get_blocked_signals` as part of the next function (where we also block some signal), and remove this one altogether?

src/signal.rs
```diff
@@ -499,4 +505,89 @@ mod tests {
...
+
+        unsafe {
+            let mut sigset: sigset_t = mem::zeroed();
+            pthread_sigmask(SIG_BLOCK, null(), &mut sigset as *mut sigset_t);
+            assert_eq!(sigismember(&sigset, SIGRTMIN() + 0), 1);
+        }
+    }
+
+    #[test]
+    fn test_unblock_signal() {
```
alexandruag: Can we merge this function with the previous one (we can maybe test unblocking the signal that we block there)?

src/signal.rs
```diff
@@ -499,4 +505,89 @@ mod tests {
...
+        assert!(killable.kill(1).is_ok());
+
+        // Pending now.
+        unsafe {
+            assert!(!SIGNAL_HANDLER_CALLED_FLAG);
+        }
+
+        clear_signal(1).unwrap();
+
+        loop {
```
alexandruag: Looks like this loop can get stuck if the flag never gets set. Does it make sense to simply wait some time period (like the current 100 ms) once, and then fail the test if the flag still didn't get set?

src/signal.rs
```diff
@@ -23,6 +26,7 @@ use std::ptr::{null, null_mut};
 use std::result;
 use std::thread::JoinHandle;
 
+/// The error cases enumeration for signal handling.
 #[derive(Debug)]
 pub enum Error {
     /// Couldn't create a sigset.
```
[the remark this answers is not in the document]
liujing2: Right, mixed usage. Thanks for pointing out. I refered to nix::sys::signal and it uses c_int. So let's keep c_int either. Does this make sense to you?

src/signal.rs
liujing2: Fixed.

src/signal.rs
liujing2: Thanks for the reviewing.
I found it is hard to define handler flags arrary/vector with SIGRTMAX() capacity, because we'd like this array as static, but calls in statics are limited to constant functions. Maybe we can only define a simple array with two members for tests usage.

src/signal.rs
liujing2: Sure. This is a good idea. I will just keep the function to be tested in unblock_signal.

src/signal.rs
liujing2: Yes, so it means the clear_signal also fails when time is over after several loop. I'll update this later.

src/signal.rs
liujing2: Thanks for your review. Addressed the comments with updated PR.

src/signal.rs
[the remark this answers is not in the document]
alexandruag: Apologies for not commenting on this particular topic earlier; I think that I would also go with `i32`, mainly because it's neater. `c_int` is a `libc` specific alias for `i32` anyway, and should this ever change it won't automatically (potentially) break our interfaces.

src/signal.rs
```diff
@@ -477,4 +478,59 @@ mod tests {
...
         // as the handle is dropped, and will be killed when the process/main thread exits.
     }
+
+    #[test]
+    fn test_block_unblock_signal() {
+        // Check if `SIGRTMIN() + 0` is blocked.
+        unsafe {
+            let mut sigset: sigset_t = mem::zeroed();
+            pthread_sigmask(SIG_BLOCK, null(), &mut sigset as *mut sigset_t);
+            assert_eq!(sigismember(&sigset, SIGRTMIN() + 0), 0);
```
alexandruag: nit: It's cleaner to have a `let signal = SIGRTMIN() + 0;` somewhere and use that name throughout the code instead of using the addition each time.

liujing2: Sure, I will keep c_int and fix the cleaner nit.

src/signal.rs
liujing2: Remove the "TODO" comment and keep enum for future.

liujing2:
@alexandruag I have checked the comments and updated PR. Thanks!

src/signal.rs
```diff
@@ -474,4 +475,61 @@ mod tests {
...
+
+        // Send a signal to the thread.
+        assert!(killable.kill(1).is_ok());
+
+        // Pending now.
+        unsafe {
+            assert!(!SIGNAL_HANDLER_CALLED[1]);
+        }
+
+        clear_signal(1).unwrap();
```
alexandruag: Just wondering, why do we call `block_signal(SIGTRMIN() + 1)`, but later just `clear_signal(1)` (without adding the `SIGTRMIN()`)?

src/signal.rs
```diff
@@ -30,7 +30,7 @@ pub enum Error {
     /// Couldn't create a sigset.
     CreateSigset(errno::Error),
     /// The wrapped signal has already been blocked.
-    SignalAlreadyBlocked(c_int),
+    SignalAlreadyBlocked(i32),
```
alexandruag: Sorry for not expressing myself clearly in the previous comment about `c_int` vs `i32` and `c_int` possibly being different on different platforms; I meant sticking with `c_int` (especially given that we apparently have stuff like `c_void` scattered around as well). At the same time, leaving the changes as they are right now in this commit does not do any harm. We can revisit the decision later if reverting them is too bothersome.

src/signal.rs
```diff
@@ -474,4 +475,61 @@ mod tests {
...
+        assert!(killable.kill(1).is_ok());
+
+        // Pending now.
+        unsafe {
+            assert!(!SIGNAL_HANDLER_CALLED[1]);
+        }
+
+        clear_signal(1).unwrap();
+
+        const MAX_WAIT_ITERS: u32 = 20;
```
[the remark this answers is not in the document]
liujing2: Sorry for replying your comments late and thanks for reviewing. Right, it seems there's enough time for the signal handler to be called, what I'm thinking is, we need ensure the tests would pass the CI every time, so the loop would help us with this. And if it indeed loops one time, it has no much impact to have the loop. So I prefer to just keep it. WDYT?

src/signal.rs
liujing2: Confusing now ;) I would like to remove the last commit in this PR(This means there will have both c_int and i32 cases), and change i32 to c_int, as my previous comment. Is this the same as your suggestion?

src/signal.rs
liujing2: Oh, this is a bug of test... And this makes me realize that, it seems hard to really block the signal in sub-thread and kill it... Currently, the block_signal actually didn't work because of not `join`. But once we `join`, the `killable` can not be used any more so can not send the signal... Do you have any suggestion here to really test the clear function with a blocked signal?

liujing2: It seems you pointed me out some problems in the test. `clear_signal` didn't report error when it finds the signal is not pending but returns due to `EAGAIN | EINTR => {}`. 
I did more experiments on the signal handling and updated the test. Per my understanding, `clear_pending()` only removes the pending signal from the list but will not trigger the signal_handler, we could instead test the `sigpending()` finally. Does this make sense?

liujing2:
> Can you also add examples for signal as well so we don't need an extra pass through this file?

Added. :)

src/signal.rs
```diff
@@ -474,4 +475,55 @@ mod tests {
...
+            }
+        });
+
+        // Send a signal to the thread.
+        assert!(killable.kill(1).is_ok());
+
+        unsafe {
+            let mut chkset: sigset_t = mem::zeroed();
+            sigpending(&mut chkset);
+            assert_eq!(sigismember(&chkset, signal), 0);
```
alexandruag: This assertion is always going to be true, because `sigpending` AFAIK returns the pending signals for the current thread, and we only send the signal to `killable`. Is the goal of this test to block a signal, then send it, check that it's pending, clear it, and finally see that it's not pending anymore? If so, can we do this sequence of operations for the current thread, without starting another one with `thread::spawn`?

src/signal.rs
```diff
@@ -474,4 +475,55 @@ mod tests {
...
+            let mut sigset: sigset_t = mem::zeroed();
+            pthread_sigmask(SIG_BLOCK, null(), &mut sigset as *mut sigset_t);
+            assert_eq!(sigismember(&sigset, signal), 0);
+        }
+
+        block_signal(signal).unwrap();
+        assert!(get_blocked_signals().unwrap().contains(&(signal)));
+
+        unblock_signal(signal).unwrap();
+        assert!(get_blocked_signals().unwrap().is_empty());
```
alexandruag: I think it might be a bit better to do an assertion like `assert!(!get_blocked_signals().unwrap().contains(&(signal)));` (in other words assert the set of blocked signals does not contain `signal`). Checking if it's totally empty might race with other tests that also block signals (like `test_clear_signal`).

src/signal.rs
liujing2: How can we send signal to current thread? :)

liujing2: Refer to http://man7.org/linux/man-pages/man2/sigpending.2.html

src/signal.rs
liujing2: Will be addressed.

src/signal.rs
alexandruag: Yea I think that works as well, and actually makes more sense than what I've said.

src/signal.rs
```diff
@@ -474,4 +484,46 @@ mod tests {
...
+        // unblocked. Pending between the time when the signal which is set as blocked
+        // is generated and when is delivered.
+        let killable = thread::spawn(move || {
+            loop {
+                // Wait for the signal being killed.
+                thread::sleep(Duration::from_millis(100));
+                if is_pending(signal) {
+                    clear_signal(signal).unwrap();
+                }
+                assert!(!is_pending(signal));
```
alexandruag: I think this assert should be moved after `clear_signal` inside the `if` block, something like this:

```rust
if is_pending(signal) {
    clear_signal(signal).unwrap();
    assert!(!is_pending(signal));
    break;
}
```
I've added the break such that the thread will exit after clearing the signal.

src/signal.rs
```diff
@@ -474,4 +484,46 @@ mod tests {
...
+            }
+        });
+
+        // Send a signal to the thread.
+        assert!(killable.kill(1).is_ok());
```
alexandruag: It looks like right now the test will kinda automatically pass because we don't wait on the thread. Coupled with the previous comment, it looks like we should join `killable` after sending the signal.

src/signal.rs
liujing2: Addressed.

src/signal.rs
liujing2: Yes, right. Updated.

src/signal.rs
```diff
@@ -382,14 +382,24 @@ mod tests {
...
     static mut SIGNAL_HANDLER_CALLED: bool = false;
 
     extern "C" fn handle_signal(_: c_int, _: *mut siginfo_t, _: *mut c_void) {
         unsafe {
+            // In the tests, there only uses vcpu signal.
             SIGNAL_HANDLER_CALLED = true;
         }
     }
 
+    fn is_pending(signal: c_int) -> bool {
```
alexandruag: This looks useful enough to become a pub function in the main `signal.rs` module. Looks good otherwise.

liujing2: Thanks for your review. I think the function is_pending would not currently been used by other codes so we could move it in the future once it’s needed. WDYT?

PR #32, opened 2019-08-14
tempfile: add support for temporary files

forbesmyester:
This commit adds support for temporary files which will compliment the
existing temporary directory functionality.

Signed-off-by: Matt Forrester <github.com@speechmarks.com>

### DIFF
--- a/src/tempfile.rs
+++ b/src/tempfile.rs
@@ -1,0 +1,167 @@
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! Struct for handling temporary files as well as any cleanup required.
+//!
+//! The temporary files will be created with a name available as well as having
+//! an exposed `fs::File` for reading/writing.
+//!
+//! The file will be removed when the object goes out of scope.
+//!
+//! # Examples
+//!
+//! ```
+//! use std::io::Write;
+//! use std::path::{Path, PathBuf};
+//! use vmm_sys_util::tempfile::TempFile;
+//! let t = TempFile::new("/tmp/testfile").unwrap();
+//! let mut f = t.as_file();
+//! f.write_all(b"hello world").unwrap();
+//! f.sync_all().unwrap();
+
+use std::ffi::{CString, OsStr};
+use std::fs;
+use std::fs::File;
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
+use std::os::unix::io::FromRawFd;
+use std::path::{Path, PathBuf};
+
+use libc;
+
+use crate::errno::{errno_result, Error, Result};
+
+/// Wrapper for working with temporary files.
+///
+/// The file will be maintained for the lifetime of the `TempFile` object.
+pub struct TempFile {
+    path: PathBuf,
+    file: File,
+}
+
+impl TempFile {
+    /// Creates the TempFile
+    ///
+    /// # Arguments
+    ///
+    /// `prefix`: The path and filename where to creat the temporary file. Six
+    /// random alphanumeric characters will be added to the end of this to form
+    /// the filename.
+    pub fn new<P: AsRef<OsStr>>(prefix: P) -> Result<TempFile> {
+        let mut os_fname = prefix.as_ref().to_os_string();
+        os_fname.push("XXXXXX");
+
+        let raw_fname = CString::new(os_fname.into_vec()).unwrap().into_raw();
+
+        // Safe because I'm sure `raw_fname` is a valid CString.
+        let fd = unsafe { libc::mkstemp(raw_fname) };
+
+        // The `raw_fname` is the same as `os_fname` (which is `prefix` with X's
+        // appended). The X's are documented as "six characters" which seem to
+        // always be alphanumerics so appears safe.
+        let c_tempname = unsafe { CString::from_raw(raw_fname) };
+        let os_tempname = OsStr::from_bytes(c_tempname.as_bytes());
+
+        if fd == -1 {
+            return errno_result();
+        }
+
+        // Safe because we checked `fd != -1` above and we uniquely own the file
+        // descriptor. This `fd` will be freed etc when `File` and thus
+        // `TempFile` goes out of scope.
+        let file = unsafe { File::from_raw_fd(fd) };
+
+        Ok(TempFile {
+            path: PathBuf::from(os_tempname),
+            file,
+        })
+    }
+
+    /// Removes the temporary file.
+    ///
+    /// Calling this is optional as dropping a `TempFile` object will also
+    /// remove the file.  Calling remove explicitly allows for better error
+    /// handling.
+    pub fn remove(&mut self) -> Result<()> {
+        fs::remove_file(&self.path).map_err(Error::from)
+    }
+
+    /// Returns the path to the tempfile if it is currently valid
+    pub fn as_path(&self) -> &Path {
+        &self.path
+    }
+
+    /// Returns a reference to the File
+    pub fn as_file(&self) -> &File {
+        &self.file
+    }
+}
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        let _ = self.remove();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+
+    #[test]
+    fn test_create_file() {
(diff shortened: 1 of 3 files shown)


src/tempfile.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// The file will be removed when the object goes out of scope.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use std::io::Write;
+    /// # use std::path::Path;
+    /// # use std::path::PathBuf;
+    /// # use vmm_sys_util::tempfile::TempFile;
+    /// # fn test_create_temp_file() -> Result<(), ()> {
```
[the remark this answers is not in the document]
forbesmyester: Can do, I was keeping consistentcy with https://github.com/rust-vmm/vmm-sys-util/blob/master/src/tempdir.rs#L32 though.

src/tempfile.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    /// Returns a reference to the File
+    pub fn as_file(&self) -> &File {
+        &self.file
+    }
+}
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        if let Some(ref p) = self.path {
```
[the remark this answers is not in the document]
forbesmyester: This is sort of copy and pasted from tempdir.rs but it is slightly better to do as you say

src/tempfile.rs
```diff
@@ -0,0 +1,142 @@
...
+        if fd == -1 {
+            return errno_result();
+        }
+
+        let temp_name = unsafe { CString::from_raw(file_bytes) };
+        let s = OsStr::from_bytes(temp_name.as_bytes());
+        let file = unsafe { File::from_raw_fd(fd) };
+
+        Ok(TempFile {
+            path: Some(PathBuf::from(s)),
```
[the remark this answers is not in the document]
forbesmyester: I thought of that too... I have no idea, but I'd be apprehensive making this API different from the API in `tempdir.rs`.

Happy to change it here if that's the view but I've not looked at the implications of changing it in `tempdir`.

src/tempfile.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+
+    #[test]
+    fn create_file() {
```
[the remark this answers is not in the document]
forbesmyester: You're right, it'd be better like that :+1:

src/tempfile.rs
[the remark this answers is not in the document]
andreeaflorescu: We are currently updating the documentation for all modules in vmm-sys-utils, and I believe tempdir was not yet updated.

src/tempfile.rs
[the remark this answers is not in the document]
andreeaflorescu: I believe we can change because it's not a breaking interface change as the fields are private anyway.

andreeaflorescu:
@forbesmyester FYI: I submitted a PR to make Tempdir use Path instead of Option<Path> https://github.com/rust-vmm/vmm-sys-util/pull/35

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// `TempFile` object will also remove the file.  Calling remove explicitly
+    /// allows for better error handling.
+    pub fn remove(&mut self) -> Result<()> {
+        if self.removed {
+            return Ok(());
+        }
+        fs::remove_file(&self.path).map_err(Error::from).map(|_| {
+            self.removed = true;
+        })
+        // match fs::remove_file(&self.path) {
```
andreeaflorescu: The commented code should be removed.

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
...
+            file,
+            removed: false,
+        })
+    }
+
+    /// Removes the temporary file.  Calling this is optional as dropping a
+    /// `TempFile` object will also remove the file.  Calling remove explicitly
+    /// allows for better error handling.
+    pub fn remove(&mut self) -> Result<()> {
+        if self.removed {
```
andreeaflorescu: A discussion about how to approach double remove also happened on this PR: https://github.com/rust-vmm/vmm-sys-util/pull/35#discussion_r317231334. I think it's better to return the Error here in case remove is called twice because that's how the `remove_file` function works as well. When called on a file that does not exist, it will return an error. I think we should just propagate that error instead of keeping an internal state with `self.removed`.

CC: @jiangliu

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
...
+use std::fs::File;
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
+use std::os::unix::io::FromRawFd;
+use std::path::{Path, PathBuf};
+
+use libc;
+
+use crate::errno::{errno_result, Result, Error};
+
+/// Create and remove a temporary file.  The file will be maintained for the
```
andreeaflorescu: The documentation should be split into summary and description. For the summary my suggestion is something like "Wrapper for working with temporary files." or something similar. The summary and description are separated by a blank line and the summary should fit on one line.

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
...
+        let file = unsafe { File::from_raw_fd(fd) };
+
+        Ok(TempFile {
+            path: PathBuf::from(os_tempname),
+            file,
+            removed: false,
+        })
+    }
+
+    /// Removes the temporary file.  Calling this is optional as dropping a
```
andreeaflorescu: Also separate this into summary and description.

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
...
+        &self.path
+    }
+
+    /// Returns a reference to the File
+    pub fn as_file(&self) -> &File {
+        &self.file
+    }
+}
+
+impl Drop for TempFile {
```
andreeaflorescu: We should also have a test for drop. I also tested this in tempdir, you can check the code in my PR: https://github.com/andreeaflorescu/vmm-sys-util/blob/7e478e5a7d4b2b5330a927b7b953780a9fd3630d/src/tempdir.rs#L127

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
+// Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! Create temporary files.
```
andreeaflorescu: nit: Can you re-word this a bit so is similar to the other documentation entries?

Some examples:
- Structures, helpers, and type definitions for working with errno.
- Enum and function for dealing with an allocated disk space by fallocate.
- Facilities for sending log message to syslog.

src/tempfile.rs
```diff
@@ -0,0 +1,167 @@
...
+//! The temporary files will be created with a name which is available as well
+//! as having an exposed `fs::File` for reading/writing.
+//!
+//! The file will be removed when the object goes out of scope.
+//!
+//! # Examples
+//!
+//! ```
+//! use std::io::Write;
+//! use std::path::Path;
```
andreeaflorescu: This can be simplified:
```use std::path::{Path, PathBuf};```

src/tempfile.rs
forbesmyester: I'll wait for that discussion to be resolved.

forbesmyester:
> Can you do a `cargo fmt`, a rebase and update the coverage as well?

fmt, yep. I'll wait on coverage till that's the only thing left as that file is destined to be permanently be in conflict and I'll just have to update it again.

src/tempfile.rs
andreeaflorescu: Looks like we chose to only write a comment that an error is returned in case the function is called twice and not keep the internal state. The tempdir PR was merged: https://github.com/rust-vmm/vmm-sys-util/pull/35

andreeaflorescu: @forbesmyester once you remove the self.remove field and just return the error from `fs::remove_file` we are ready to merge this PR.

forbesmyester: I only have one branch in my code (that I don't think I can reliably execute). Any thoughts how I can get the 0.1% test coverage back?

andreeaflorescu: I just ran the coverage test locally on your branch + upstream changes and it looks like the coverage is increased from to 74.9. I also looked at the covered code and it seems good. Just do a rebase and add the appropriate coverage in coverage_config.json. We can merge this PR next.

forbesmyester: Thanks

forbesmyester:
Good now I think...

PR #34, opened 2019-08-20
poll: documentation improvements

liujing2:


liujing2:
> We also need examples for the public interface.

Will add it.

src/poll.rs
```diff
@@ -129,30 +135,38 @@ impl PollToken for () {
...
+    ///
+    /// True if the `fd` associated with this token in
+    /// [`PollContext::add`](struct.PollContext.html#method.add) has been hungup on.
     pub fn hungup(&self) -> bool {
         self.event.events & (EPOLLHUP as u32) != 0
     }
 }
 
-/// An iterator over some (sub)set of events returned by `PollContext::wait`.
+/// An iterator over some (sub)set of events returned by
```
andreeaflorescu: nit: An iterator over a subset ...

liujing2: Addressed.

PR #35, opened 2019-08-23
Tempdir Documentation & Refactoring

andreeaflorescu:
- added missing documentation
- refactored existing documentation examples so that we don't use functions
- the path field is now not optional anymore as there was no good reason for it to be optional

### DIFF
--- a/src/tempdir.rs
+++ b/src/tempdir.rs
@@ -4,6 +4,8 @@
 //
 // SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
 
+//! Structure for handling temporary directories.
+
 use libc;
 use std::ffi::{CString, OsStr, OsString};
 use std::fs;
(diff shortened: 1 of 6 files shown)


src/tempdir.rs
```diff
@@ -54,30 +50,49 @@ impl TempDir {
...
+    /// # Examples
+    ///
+    /// ```
+    /// # use std::path::Path;
+    /// # use std::path::PathBuf;
+    /// # use vmm_sys_util::tempdir::TempDir;
+    /// let temp_dir = TempDir::new("/tmp/testdir").unwrap();
+    /// temp_dir.remove().unwrap();
+    ///
+    pub fn remove(&self) -> Result<()> {
```
jiangliu: Should we use some marker to store the state whether the tempdir has already been deleted to avoid duplicated deleting?

andreeaflorescu: I think the function will return error in the case where remove is called twice. I would rather add a documentation entry specifying it should only be called once.

PR #36, opened 2019-08-26
seek_hole: update documentation

andreeaflorescu:
Added high level documentation. Separated existing documentation into
summary and details.

This commit does not include examples for SeekHole trait because we
need a lot of boiler plate code to do that (we need to create a
temporary directory, create a file, write something to the file
and only then we are able to call into seek_data or seek_hole).

### DIFF
--- a/src/seek_hole.rs
+++ b/src/seek_hole.rs
@@ -4,6 +4,8 @@
 //
 // SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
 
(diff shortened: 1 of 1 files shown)


liujing2:
Need a refactor before merging. Thanks!

andreeaflorescu:
@liujing2 what refactor is needed?

liujing2:
> @liujing2 what refactor is needed?

Typo. Just rebase. Codes are fine to me. :)

PR #37, opened 2019-08-26
Fix fam module documentation

serban300:


src/fam.rs
```diff
@@ -5,6 +5,18 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
```
liujing2: Could you also improve the license file name here? Thanks!

serban300: Sure. Done.

Issue #28, opened 2019-08-01
syslog - cleanup


andreeaflorescu:
The syslog implementation and comments need to be cleaned up. There are a lot of "legacy" comments, TODOs and links added by the CrosVM team that are related to CrosVM and should not be in vmm-sys-util.
- https://github.com/rust-vmm/vmm-sys-util/blob/b546b9d068a22ac866917516db8623b0dbeef831/src/syslog.rs#L164
- https://github.com/rust-vmm/vmm-sys-util/blob/b546b9d068a22ac866917516db8623b0dbeef831/src/syslog.rs#L170

I also believe that the code in syslog is too complicated for no good reasons. While this makes sense in the CrosVM context, I think we should find a better workaround as I consider it to be relatively unsafe. Specifically the part that "guesses" that syslog is the first opened fd.
https://github.com/rust-vmm/vmm-sys-util/blob/b546b9d068a22ac866917516db8623b0dbeef831/src/syslog.rs#L181

There are some things that can be simplified, the issue needs to be better scoped.

andreeaflorescu:
Closing this as we removed the syslog for the time being.

Issue #23, opened 2019-07-31
Check results of calling syslog function


andreeaflorescu:
Most tests in the syslog are only calling functions/macros without actually checking that the functions/macros behaved as expected.

We should add checks for the expected entries in the syslog files.

andreeaflorescu:
Closing this as syslog was temporarily removed. When re-added the unit tests should be checked to see if they actually validate the result of calling one function.

PR #39, opened 2019-08-29
removed syslog

andreeaflorescu:
The syslog module needs serious refactoring before we can publish it
on crates.io. This is blocking publishing vmm-sys-utils, so until
we get more time and we have an explicit request for this crate, we
are temporarily removing it.

Related issues:
- https://github.com/rust-vmm/vmm-sys-util/issues/28
- https://github.com/rust-vmm/vmm-sys-util/issues/23


sameo:
crosvm uses it, but by the time they may switch to this crate, we'll have that one back. LGTM.

andreeaflorescu:
@sameo When we last talked to CrosVM they said that they'll create a wrapper for this crate and add all the other functionality that they need in the wrapper. I wouldn't see it as a problem.

PR #42, opened 2019-09-17
Implement From<Vec<T>> for FamStructWrapper<T>

jiangliu:
Implement From<Vec<T>> for FamStructWrapper, so we could move the Vec
instead of copy its content. Also implement trait From<&Vec<T>> for
convenience.

Signed-off-by: Jing Liu <jing2.liu@linux.intel.com>

### DIFF
--- a/src/fam.rs
+++ b/src/fam.rs
@@ -398,6 +398,20 @@ impl<T: Default + FamStruct + Clone> Clone for FamStructWrapper<T> {
     }
 }
 
+impl<T: Default + FamStruct + Clone> From<Vec<T>> for FamStructWrapper<T> {
+    fn from(vec: Vec<T>) -> Self {
+        FamStructWrapper { mem_allocator: vec }
+    }
+}
+
+impl<T: Default + FamStruct + Clone> From<&Vec<T>> for FamStructWrapper<T> {
+    fn from(vec: &Vec<T>) -> Self {
+        FamStructWrapper {
+            mem_allocator: vec.clone(),
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


sameo:
@jiangliu Please change `coverage_config.json` from 83.5 to 83.7 to make the CI pass...

jiangliu:
> @jiangliu Please change `coverage_config.json` from 83.5 to 83.7 to make the CI pass...

Thanks, fixed:)

src/fam.rs
```diff
@@ -398,6 +398,20 @@ impl<T: Default + FamStruct + Clone> Clone for FamStructWrapper<T> {
     }
 }
 
+impl<T: Default + FamStruct + Clone> From<Vec<T>> for FamStructWrapper<T> {
```
serban300: This doesn't seem right. The result definetly won't be a concatenation of all the elements of all the FamStruct instances in the supplied Vec. What is the desired behaviour ?

jiangliu: The usage scenario is for serialization/deserialization. The underline memory layout is ready for FamStructWrapper after deserialization, so we could skip the memcopy copy.

jiangliu: Please refer to line 42 at https://github.com/jiangliu/vmm-serde/blob/v1/src/ffi.rs
Good catch! The implementation of `impl<T: Default + FamStruct + Clone> From<&Vec<T>> for FamStructWrapper<T>`  depends on the implementation of T::Entry::Clone, we should enforce 'T: Copy` instead of `T: Clone`.

alexandruag: Doing this via implementing `From<Vec<T>>` as opposed to having a specific function/constructor with a more descriptive name (such as `from_allocator` etc.) can be a bit confusing.

Also, looks like we could get a serialization discussion in `rust-vmm` going since this will probably become a very important topic, especially as we think about adding more building blocks. Should we use https://github.com/rust-vmm/community/issues/76 for that?

jiangliu: So how about reimplementing the From<Vec<T>> as FamStructWrapper::from_vector()?
I have confirmed that the `impl From<&Vec<T>> for FamStructWrapper<T>` has the same assumption as impl<T: Default + FamStruct + Clone> Clone for FamStructWrapper<T>, so it should be OK.

jiangliu: @alexandruag we do need some discussion and cooperation on serialization/deserialization to build a common infrastructure for the rust-vmm project,  and other crates should support serde by default.

serban300: I think `from_vector` still might be confusing. `from_mem_allocator` would be closest to reality. Also a comment on what it does and why it is needed would be super helpful. 

Maybe also the unit test could be adapted to reflect the usage scenario better: for example having a u32 slice, transmuting it to a Vec<MockFamStruct> if possible, calling `from_mem_allocator` and checking that the elements are as expected.

PR #44, opened 2019-10-07
Add CHANGELOG

sameo:
Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>


jiangliu:
There are two more modules: sock_ctrl_msg and aio:)

sameo:
On Mon, Oct 07, 2019 at 04:31:05PM -0700, Jiang Liu wrote:
> There are two more modules: sock_ctrl_msg and aio:)

Good call. See PR #45.

PR #46, opened 2019-10-08
Post release fixes

sameo:
* clippy warning fixes
* `Cargo.toml` license

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ description = "A system utility set"
 repository = "https://github.com/rust-vmm/vmm-sys-util"
 readme = "README.md"
 keywords = ["utils"]
-license = "Apache-2.0 OR BSD-3-Clause"
+license = "Apache-2.0 AND BSD-3-Clause"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


andreeaflorescu:
@sameo why aren't the clippy warnings generating errors? Should we update the CI as well?

Cargo.toml
```diff
@@ -6,7 +6,7 @@ description = "A system utility set"
 repository = "https://github.com/rust-vmm/vmm-sys-util"
 readme = "README.md"
 keywords = ["utils"]
-license = "Apache-2.0 OR BSD-3-Clause"
+license = "Apache-2.0 AND BSD-3-Clause"
```
andreeaflorescu: What about we also update the version to 0.1.1 and then publish the new version on crates.io?

sameo:
On Tue, Oct 08, 2019 at 02:29:31AM -0700, Andreea Florescu wrote:
> @sameo why aren't the clippy warnings generating errors? Should we update the CI as well?

I think the container rust version does not catch those yet. I was
seeing those warnings on my 1.37 toolchain.

PR #50, opened 2019-10-22
cargo: Bump version to 0.2.0

acatangiu:
Release an incremental version to include the new FAM functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
+# v0.2.0
+
+* fam: updated the macro that generates implementions of FamStruct to
+  also take a parameter that specifies the name of the flexible array
+  member.
+
 # v0.1.1
 
 * Fixed the Cargo.toml license.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
+# v0.1.2
```
aghecenco: nit: the commit message and the PR title say the bump is to v0.1.1 instead

CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
...
+
+* fam: updated the macro that generates implementions of FamStruct to
```
andreeaflorescu: Even though this is a tiny change, I think we might need to use 0.2.0 instead because it is a breaking change.

acatangiu:
@aghecenco @andreeaflorescu apparently I can muck up many things when doing a simple version bump 🤣 . Good catch, fixed.

PR #54, opened 2019-11-20
Block signals handlers when registering a new signal handler

### LINKED ISSUE #52: Block other signals while a registered signal handler is running
Just noticed that `register_signal_handler` from `signal.rs`: https://github.com/rust-vmm/vmm-sys-util/blob/472614c3a21231f6a0c3696fab83fb34d6592ade/src/signal.rs#L191, lacks a way to specify to a signal handler what signals to be blocked while it is running.

To be specific, I am referring to a way to initialize the `sa_mask` field of a `struct sigaction`, when calling `register_signal_handler`. An example of my usecase is here [0].

Are you fond of a way to achieve this with the current implementation?
If it couldn't be achieved with the current implementation, I can gladly contribute with a PR.

[0] https://www.gnu.org/software/libc/manual/html_node/Blocking-for-Handler.html#Blocking-for-Handler.

iulianbarbu:
Fixes #52 

Blocked all signals handlers execution when registering
a signal handler, for the registered signal handler.
Refactored SignalHandler enum to a SignalHandler type.

Signed-off-by: Iulian Barbu <iul@amazon.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# v0.3.0
+
+* Removed `for_vcpu` argument from `signal::register_signal_handler` and
+`signal::validate_signal_num`. Users can now pass absolute values for all valid 
+signal numbers.
+* Removed `flag` argument of `signal::register_signal_handler` public methods,
+which now defaults to `libc::SA_SIGINFO`.
+
 # v0.2.1
 
 * Fixed the FamStructWrapper Clone implementation to avoid UB.
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -11,8 +11,8 @@
 
 use libc::{
     c_int, c_void, pthread_kill, pthread_sigmask, pthread_t, sigaction, sigaddset, sigemptyset,
-    siginfo_t, sigismember, sigpending, sigset_t, sigtimedwait, timespec, EAGAIN, EINTR, EINVAL,
-    SIGHUP, SIGSYS, SIG_BLOCK, SIG_UNBLOCK,
+    sigfillset, siginfo_t, sigismember, sigpending, sigset_t, sigtimedwait, timespec, EAGAIN,
+    EINTR, EINVAL, SIG_BLOCK, SIG_UNBLOCK,
 };
 
 use errno;
@@ -80,35 +80,11 @@ impl Display for Error {
 /// A simplified [Result](https://doc.rust-lang.org/std/result/enum.Result.html) type
 /// for operations that can return [`Error`](Enum.error.html).
 pub type SignalResult<T> = result::Result<T, Error>;
-type SiginfoHandler = extern "C" fn(num: c_int, info: *mut siginfo_t, _unused: *mut c_void) -> ();
 
-/// Enum of signal handlers for
+/// Public alias for a signal handler.
 /// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
-pub enum SignalHandler {
-    /// Corresponds to the `sa_sigaction` member of
-    /// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html).
-    Siginfo(SiginfoHandler),
-}
-
-impl SignalHandler {
-    fn set_flags(act: &mut sigaction, flag: c_int) {
-        act.sa_flags = flag;
-    }
-}
-
-/// Convert a [`SignalHandler`](enum.SignalHandler.html) into a
-/// [`sigaction`](http://man7.org/linux/man-pages/man2/sigaction.2.html)
-impl Into<sigaction> for SignalHandler {
-    fn into(self) -> sigaction {
-        let mut act: sigaction = unsafe { mem::zeroed() };
-        match self {
-            SignalHandler::Siginfo(function) => {
-                act.sa_sigaction = function as *const () as usize;
-            }
-        }
-        act
-    }
-}
+pub type SignalHandler =
+    extern "C" fn(num: c_int, info: *mut siginfo_t, _unused: *mut c_void) -> ();
 
 extern "C" {
     fn __libc_current_sigrtmin() -> c_int;
@@ -129,14 +105,12 @@ fn SIGRTMAX() -> c_int {
 
 /// Verify that a signal number is valid.
 ///
-/// VCPU signals must be enclosed within the OS
-/// limits for realtime signals. The remaining ones need
-/// to be between the minimum (`SIGHUP`) and maximum (`SIGSYS`) values.
+/// Supported signals range from `SIGHUP` to `SIGSYS` and from `SIGRTMIN` to `SIGRTMAX`.
+/// We recommend using realtime signals `[SIGRTMIN, SIGRTMAX]` for VCPU threads.
 ///
 /// # Arguments
 ///
-/// * `num`: the signal number to be verify.
-/// * `for_vcpu`: if `num` is a vcpu signal.
+/// * `num`: the signal number to be verified.
 ///
 /// # Examples
 ///
@@ -144,18 +118,14 @@ fn SIGRTMAX() -> c_int {
 /// extern crate vmm_sys_util;
 /// use vmm_sys_util::signal::validate_signal_num;
 ///
-/// let num = validate_signal_num(1, true).unwrap();
+/// let num = validate_signal_num(1).unwrap();
 /// ```
-pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
-    if for_vcpu {
-        let actual_num = num + SIGRTMIN();
-        if actual_num <= SIGRTMAX() {
-            return Ok(actual_num);
-        }
-    } else if SIGHUP <= num && num <= SIGSYS {
-        return Ok(num);
+pub fn validate_signal_num(num: c_int) -> errno::Result<()> {
+    if (libc::SIGHUP <= num && num <= libc::SIGSYS) || (SIGRTMIN() <= num && num <= SIGRTMAX()) {
+        Ok(())
+    } else {
+        Err(errno::Error::new(EINVAL))
     }
-    Err(errno::Error::new(EINVAL))
 }
 
 /// Register the signal handler of `signum`.
@@ -170,9 +140,6 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
 ///
 /// * `num`: the signal number to be registered.
 /// * `handler`: the signal handler function to register.
-/// * `for_vcpu`: whether `num` is a vcpu signal.
-/// * `flag`: specify the behavior of the signal.
-///    Set SA_SIGINFO or SA_RESTART if wants to restart after signal received.
 ///
 /// # Examples
 ///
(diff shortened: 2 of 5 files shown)


acatangiu:
I'm not sure there is a usecase to block specific signals while in a signal handler. I would just block all and explicitly declare that behaviour in the function documentation.

Re-entrant or cascaded signal handlers are usually discouraged.

iulianbarbu:
> I'm not sure there is a usecase to block specific signals while in a signal handler. I would just block all and explicitly declare that behaviour in the function documentation.
> 
> Re-entrant or cascaded signal handlers are usually discouraged.

I am fine with that. I will wait for more feedback though, then make the changes.

iulianbarbu:
@acatangiu , I've addressed your concerns. Please take a look: https://github.com/rust-vmm/vmm-sys-util/commit/8cd664d096cb2e346f03c0999e9f8e9a2ecfd2bf.

src/signal.rs
```diff
@@ -183,21 +159,30 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
...
 ) -> errno::Result<()> {
     let num = validate_signal_num(num, for_vcpu)?;
-    let mut act: sigaction = handler.into();
-    SignalHandler::set_flags(&mut act, flag);
-    match sigaction(num, &act, null_mut()) {
+
+    // Safe, because this is a POD struct.
+    let mut act: sigaction = unsafe { mem::zeroed() };
+    act.sa_sigaction = handler as *const () as usize;
+    act.sa_flags = flag;
```
acatangiu: When using `sa_sigaction` instead of `sa_handler` to specify the handler, `sa_flags` needs to have `libc::SA_SIGINFO` set.

So at a minimum this line should be:
```rust
act.sa_flags = flag | libc::SA_SIGINFO
```

Looking at the other possible values of `sa_flags` I suggest we simply don't expose this as none of them seem like something VM's should use.

My preference is removing `flag` altogether from this function's param list and just set `sa_flags = SA_SIGINFO`.

src/signal.rs
```diff
@@ -183,21 +159,30 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
...
+
+    // Block all signals while the `handler` is running.
+    // This makes async signal operations safe, because
+    // execution of signal handlers can not overlap.
+    if unsafe { sigfillset(&mut act.sa_mask as *mut sigset_t) } < 0 {
+        return errno::errno_result();
+    }
+
+    match unsafe { sigaction(num, &act, null_mut()) } {
```
acatangiu: nit: add a comment explaining why this is safe.

src/signal.rs
```diff
@@ -183,21 +159,30 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
...
-///     register_signal_handler(0, SignalHandler::Siginfo(handle_signal), true, SA_SIGINFO);
-/// }
+/// register_signal_handler(0, handle_signal, true, SA_SIGINFO);
 /// ```
 
-pub unsafe fn register_signal_handler(
+pub fn register_signal_handler(
     num: c_int,
     handler: SignalHandler,
     for_vcpu: bool,
```
acatangiu: I really think this `for_vcpu` is simply wrong. Like I said before, let's just remove it. You can do it in this PR or I can open a new one after, whichever you like.

acatangiu:
nit: suggest change in commit summary from:
```
- signal: block all sig hndlrs at sig handler regist
+ signal: block other signals during signal handling
```

acatangiu:
Also add entries to the `CHANGELOG` pertaining to the API changes.

iulianbarbu:
@acatangiu , I think I've included the feedback from the last review. Please take a look: https://github.com/rust-vmm/vmm-sys-util/commit/ab291fc935db89b94dd8d792952bfce7d14a81e3.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v0.3.0
+
+* Removed `for_vcpu` args from `signal::register_signal_handler` & from
+`signal::validate_signal_num`, which puts inside the same bucket what
+we called `VCPU signals` and the others.
+* Removed `flag` arg of `signal::register_signal_handler` public methods,
+which defaults to `libc::SA_SIGINFO`.
```
acatangiu:
```suggestion
which now defaults to `libc::SA_SIGINFO`.
```

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v0.3.0
+
+* Removed `for_vcpu` args from `signal::register_signal_handler` & from
+`signal::validate_signal_num`, which puts inside the same bucket what
+we called `VCPU signals` and the others.
```
acatangiu: I'd rephrase this to:
```
Removed `for_vcpu` argument from `signal::register_signal_handler` and
`signal::validate_signal_num`. Users now have full control which signals can
be used for their threads.
```

src/signal.rs
```diff
@@ -129,32 +99,27 @@ fn SIGRTMAX() -> c_int {
 
 /// Verify that a signal number is valid.
 ///
-/// VCPU signals must be enclosed within the OS
-/// limits for realtime signals. The remaining ones need
+/// Usually, VCPU signals are assigned and must be enclosed within the OS
+/// limits for realtime signals, while the remaining ones need
 /// to be between the minimum (`SIGHUP`) and maximum (`SIGSYS`) values.
```
acatangiu: **must be enclosed** is incorrect.

Let's make the whole comment clearer:
```rust
/// Supported signals range from `SIGHUP` to `SIGUSR2` and from `SIGRTMIN` to `SIGRTMAX`.
/// We recommend using realtime signals `[SIGRTMIN, SIGRTMAX]` for VCPU threads.
```

src/signal.rs
```diff
@@ -129,32 +99,27 @@ fn SIGRTMAX() -> c_int {
...
 /// ```
-pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
-    if for_vcpu {
-        let actual_num = num + SIGRTMIN();
-        if actual_num <= SIGRTMAX() {
-            return Ok(actual_num);
-        }
-    } else if SIGHUP <= num && num <= SIGSYS {
+pub fn validate_signal_num(num: c_int) -> errno::Result<c_int> {
+    if SIGHUP <= num && num <= SIGRTMAX() {
```
acatangiu: The `[SIGRTMIN, SIGRTMAX]` range varies depending on arch and libc implementation. There might be a gap between the 32 standard signal numbers and the 32 real-time signal numbers.

```suggestion
    if (SIGHUP <= num && num <= SIGUSR2) || (SIGRTMIN() <= num && num <= SIGRTMAX()) {
```

src/signal.rs
```diff
@@ -538,22 +489,20 @@ mod tests {
...
-            register_signal_handler(0, SignalHandler::Siginfo(handle_signal), true, SA_SIGINFO)
-                .expect("failed to register vcpu signal handler");
-        }
+        register_signal_handler(SIGRTMIN(), handle_signal)
+            .expect("failed to register vcpu signal handler");
 
         let killable = thread::spawn(|| loop {});
 
-        let res = killable.kill(SIGRTMAX());
+        let res = killable.kill(SIGRTMAX() + 1);
```
acatangiu: Is `SIGRTMAX()` usable? As in the first unusable is `SIGRTMAX()+1`?

iulianbarbu: Hadn't found any resource that discourages usage of SIGRTMIN, respectively of SIGRTMAX, as signums for registration of signal handlers. The [signal(7)](http://man7.org/linux/man-pages/man7/signal.7.html) doc does not discourage the usage of SIGRTMIN, or SIGRTMAX, but expresses the need to use checks to avoid the case when `SIGRTMIN+n` exceeds `SIGRTMAX`.

As a seprate note, at this specific line, we have to use a signum which exceeds `SIGRTMAX()` because we look for the error thrown from `validate_signal_num` function.

This line could've been in the past as simple as:
 ```suggestion
        let res = killable.kill(16);
```

`SIGRTMAX()` usage, as an argument, is superfluous and does not mean that `killable.kill()` call will explicitly raise the `SIGRTMAX()` signal for the specific thread, but instead it would've raised the signal for the signum `SIGRTMIN() + SIGRTMAX()`, which is not a valid signal number also. While in the new case, with the simplifications from this PR, it is necessary to make the call for `killable.kill()` with an argument which explicitly exceeds `SIGRTMAX()`, so that is why I chose `SIGRTMAX() + 1`.

File: src/signal.rs (hunk 6 above)
Comment (iulianbarbu) [reply src/signal.rs]: 👍 done that.

File: src/signal.rs [hunk 9]
```diff
@@ -129,32 +105,26 @@ fn SIGRTMAX() -> c_int {
 
 /// Verify that a signal number is valid.
 ///
-/// VCPU signals must be enclosed within the OS
-/// limits for realtime signals. The remaining ones need
-/// to be between the minimum (`SIGHUP`) and maximum (`SIGSYS`) values.
+/// Supported signals range from `SIGHUP` to `SIGUSR2` and from `SIGRTMIN` to `SIGRTMAX`.
```
acatangiu:
```suggestion
/// Supported signals range from `SIGHUP` to `SIGSYS` and from `SIGRTMIN` to `SIGRTMAX`.
```

src/signal.rs
```diff
@@ -129,32 +105,26 @@ fn SIGRTMAX() -> c_int {
...
+/// let num = validate_signal_num(1).unwrap();
 /// ```
-pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
-    if for_vcpu {
-        let actual_num = num + SIGRTMIN();
-        if actual_num <= SIGRTMAX() {
-            return Ok(actual_num);
-        }
-    } else if SIGHUP <= num && num <= SIGSYS {
+pub fn validate_signal_num(num: c_int) -> errno::Result<c_int> {
```
acatangiu: Sorry for the late feedback, but since this function is now just validating and not mutating the signal number, how about we change it to returning a `bool` or a `Result<()>` (I prefer the latter):

```suggestion
pub fn validate_signal_num(num: c_int) -> errno::Result<()> {
```

iulianbarbu: Yep, sure. I chose to return `Result<()>` because it is compatible with unit tests, and could be promptly returned in case of `Error`, while returning `bool` needs writing checks and explicit return of Error if the function call returns `false`.

acatangiu:
@sameo @jiangliu @andreeaflorescu please review.

src/signal.rs
```diff
@@ -129,32 +105,26 @@ fn SIGRTMAX() -> c_int {
...
 ///
-/// VCPU signals must be enclosed within the OS
-/// limits for realtime signals. The remaining ones need
-/// to be between the minimum (`SIGHUP`) and maximum (`SIGSYS`) values.
+/// Supported signals range from `SIGHUP` to `SIGSYS` and from `SIGRTMIN` to `SIGRTMAX`.
+/// We recommend using realtime signals `[SIGRTMIN, SIGRTMAX]` for VCPU threads.
 ///
 /// # Arguments
 ///
 /// * `num`: the signal number to be verify.
```
aghecenco: nit
```suggestion
 /// * `num`: the signal number to be verified.
```

src/signal.rs
```diff
@@ -183,21 +150,26 @@ pub fn validate_signal_num(num: c_int, for_vcpu: bool) -> errno::Result<c_int> {
...
+pub fn register_signal_handler(num: c_int, handler: SignalHandler) -> errno::Result<()> {
+    validate_signal_num(num)?;
+
+    // Safe, because this is a POD struct.
+    let mut act: sigaction = unsafe { mem::zeroed() };
+    act.sa_sigaction = handler as *const () as usize;
+    act.sa_flags = libc::SA_SIGINFO;
+
+    // Block all signals while the `handler` is running.
+    // This makes async signal operations safe, because
```
aghecenco: nit: maybe explain that blocking other signals is needed to make sure the execution of the handler continues uninterrupted if another signal comes

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v0.3.0
+
+* Removed `for_vcpu` argument from `signal::register_signal_handler` and
+`signal::validate_signal_num`. Users now have full control which signals can
+be used for their threads.
+* Removed `flag` arg of `signal::register_signal_handler` public methods,
```
aghecenco: nit: you use `argument` above and `arg` here, please make them consistent

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# v0.3.0
+
+* Removed `for_vcpu` argument from `signal::register_signal_handler` and
+`signal::validate_signal_num`. Users now have full control which signals can
+be used for their threads.
```
aghecenco: The users had full control before this change too, the syntax was just different:
```suggestion
`signal::validate_signal_num`. Users can now pass absolute values for all valid signal numbers.
```

acatangiu: They didn't actually. At least not without "cheating". In the previous version, setting "for_vcpu" to `true` would restrict signal numbers to [SIGRTMIN, SIGRTMAX]. You could argue that one could use the function for setting up a signal handler for a vcpu but set `for_vcpu` to `false` but that's "cheating" 😛 

Either way, I'm not attached to this wording, I'm fine to change it, just wanted to give more context.

iulianbarbu:
@aghecenco , I've addressed all of your comments. Please take a look again at: https://github.com/rust-vmm/vmm-sys-util/commit/96ec93830ebf0c4ea3fb30f29ff4f3b93efed4bc.

acatangiu:
@rust-vmm/gatekeepers please review

PR #55, opened 2019-11-20
fam: fix Clone for FamStructWrapper - bump version to include the fix

acatangiu:
FamStructWrapper should not require `Clone` for the fam-struct `T` it safely wraps.
The underlying unsafe fam-struct type `T` cannot safely implement `Clone`. Attempting to reference the fam-member of such a clone can lead to UB.

We can still provide a safe `Clone` for the safe `FamStructWrapper`: instead of cloning the underlying `Vec<T>`, we can simply build a new `FamStructWrapper` using the safe builder starting from the original `[T::Entry]` where we know that `T::Entry` implements `Copy`.

Also bump the version to 0.2.1 to publish a fixed version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+.idea
 **/*.rs.bk
 Cargo.lock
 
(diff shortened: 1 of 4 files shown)


acatangiu:
CC @serban300

PR #56, opened 2019-12-02
tmpfile and tmpdir creation inside /tmp

iulianbarbu:
Signed-off-by: Iulian Barbu <iul@amazon.com>

Implemented new methods for TempFile struct, respectively for TempDir struct:
- TempFile/TempDir::new() creates a new temp file inside $TMPDIR if set, otherwise inside /tmp.
- TempFile/TempDir::new_in(path) creates a new temp file inside an already existing valid path.
- TempFile/TempDir::new_with_prefix(prefix) is the old TempFile/TempDir::new(prefix) method.

### DIFF
--- a/src/tempdir.rs
+++ b/src/tempdir.rs
@@ -5,12 +5,13 @@
 // SPDX-License-Identifier: (Apache-2.0 AND BSD-3-Clause)
 
 //! Structure for handling temporary directories.
-
-use libc;
+use std::env::temp_dir;
 use std::ffi::{CString, OsStr, OsString};
 use std::fs;
 use std::os::unix::ffi::OsStringExt;
 use std::path::{Path, PathBuf};
+
+use libc;
 
 use crate::errno::{errno_result, Error, Result};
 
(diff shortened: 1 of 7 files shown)


jiangliu:
There are some intelj related files included, please help to remove them.

src/tempdir.rs
```diff
@@ -5,13 +5,14 @@
...
 //! Structure for handling temporary directories.
-
-use libc;
+use std::env::temp_dir;
 use std::ffi::{CString, OsStr, OsString};
 use std::fs;
 use std::os::unix::ffi::OsStringExt;
 use std::path::{Path, PathBuf};
 
+use libc;
```
andreeaflorescu: nit: why not have this where it previously was?

iulianbarbu: I structured the `use`s in the following groups: `std` usages in one group, other libs usages in another, same crate modules usages respectively in another.

iulianbarbu:
> There are some intelj related files included, please help to remove them.

@jiangliu , removed the intelj files. 

> Please also add more descriptive commit message.

@andreeaflorescu , updated.

Please take a look [here](https://github.com/rust-vmm/vmm-sys-util/commit/3b94a8b57d1d13babb3965a9990e45143ab62129).

PR #60, opened 2019-12-10
tempfile: implement relevant File traits

dianpopa:
Propagate Read, Write, Flush, Seek for
TempFile which are essential traits when
dealing with File objects.

Signed-off-by: Diana Popa <dpopa@amazon.com>

### DIFF
--- a/src/tempfile.rs
+++ b/src/tempfile.rs
@@ -26,6 +26,7 @@ use std::env::temp_dir;
 use std::ffi::{CString, OsStr};
 use std::fs;
 use std::fs::File;
+use std::io::{self, Read, Seek, SeekFrom, Write};
 use std::os::unix::ffi::{OsStrExt, OsStringExt};
 use std::os::unix::io::FromRawFd;
 use std::path::{Path, PathBuf};
@@ -128,6 +129,28 @@ impl TempFile {
 impl Drop for TempFile {
     fn drop(&mut self) {
         let _ = self.remove();
+    }
+}
+
+impl Read for TempFile {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.file.read(buf)
+    }
+}
+
+impl Write for TempFile {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.file.write(buf)
+    }
+    #[inline]
+    fn flush(&mut self) -> io::Result<()> {
+        self.file.flush()
+    }
+}
+
+impl Seek for TempFile {
+    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        self.file.seek(pos)
     }
 }
 


dianpopa:
> Tempfile was thought to be just a way to get a unique path. The name is not very suggestive to that end goal unfortunately.

 So, you are saying that the file inside the TempFile was not meant to be used? I think this is quite a big miss.  You obtain a Tempfile object but you cannot actually use it to do any of the usual operations that you would do on a File even though TempFile has a File in it. In that case, how should the file inside TempFile be used? Calling `as_file` gives you a reference but what to do when you need an owned File?

> If we want to move `Tempfile` to a design where it is used like a regular File, then we should also have a builder that can create files with different permissions. Right now Tempfile creates a dummy File with read only permissions. In this context, implementing the `Write` trait doesn't really make sense because it would never work without changing the metadata of the file before calling the `write` method.

Why would you need a builder for being able to change files permissions? Why not call [set_permissions](https://doc.rust-lang.org/std/fs/struct.File.html#method.set_permissions) on the File? Also, are you implying that people using tempfile right now from vmm-sys-util should only be using it if they need a read-only File?

andreeaflorescu:
> > Tempfile was thought to be just a way to get a unique path. The name is not very suggestive to that end goal unfortunately.
> 
> So, you are saying that the file inside the TempFile was not meant to be used? I think this is quite a big miss. You obtain a Tempfile object but you cannot actually use it to do any of the usual operations that you would do on a File even though TempFile has a File in it. In that case, how should the file inside TempFile be used? Calling `as_file` gives you a reference but what to do when you need an owned File?

TempFile as it is implemented right now cannot really give ownership of the inner file because it needs the ownership to remove the file on `Drop`. We can change this functionality, but the important part is to still get the temporary files.

> > If we want to move `Tempfile` to a design where it is used like a regular File, then we should also have a builder that can create files with different permissions. Right now Tempfile creates a dummy File with read only permissions. In this context, implementing the `Write` trait doesn't really make sense because it would never work without changing the metadata of the file before calling the `write` method.
> 
> Why would you need a builder for being able to change files permissions? Why not call [set_permissions](https://doc.rust-lang.org/std/fs/struct.File.html#method.set_permissions) on the File? Also, are you implying that people using tempfile right now from vmm-sys-util should only be using it if they need a read-only File?

No, I am not saying that tempfile should be used only for read-only files. I am saying that right now the caller is expected to change the permissions and essentially use TempFile more like a path generator that removes the file on drop.

A tempfile builder made sense to me because you would configure the type of file from the beginning and then you wouldn't need to work with the inner tempfile types (such as File) to change the permissions. The builder could have the following methods:
- `with_prefix` (this would create a new tempfile with a given prefix and replace the existing new_with_prefix)
- `read`
- `write`
- `open`

As I said before, this is just one of the options. Which one we pick is not important. What I think is important is to offer one clear way of working with TempFiles. If we want `TempFile` to behave as a File, than we should also offer the common functionality of setting permissions directly on the `TempFile` object, and not relying on this functionality being offered by the inner file type.

sameo:
Right now we are creating these temp files with read and write permissions to the owner. That sounds quite reasonable and typical for a temporary file, so what are the actual usage patterns for allowing the caller to set read-only or write-only permissions on a temp file? There is a glibc `mkostemp()` extension but a) it's not even POSIX and b) it's mostly for appending, which to me is really a corner case.

So why would we make this API more complex by allowing the caller to set unusual file permissions on a temp file?

andreeaflorescu:
Closing this as the problem seems to be fixed by the already merged PR #70. Please re-open if that's not the case.

Issue #65, opened 2019-12-18
Add serialization support to FamStruct


aghecenco:
`CpuId` and `Msrs` will need to be serialized for https://github.com/firecracker-microvm/firecracker/issues/1184. The FAM wrappers need to be (conditionally) enhanced with serde support.

PR #66, opened 2019-12-18
fam wrappers: add conditional serde compatibility

aghecenco:
#65

This PR makes `FamStructWrapper`s `serde`-serializable, provided that the wrapped data type is also `serde`-serializable.
The functionality is gated by the `with-serde` compilation feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Added conditionally compiled `serde` compatibility to `FamStructWrapper`,
+  gated by the `with-serde` feature.
+
 # v0.4.0
 
 * Added Windows support for TempFile and errno::Error.
@@ -49,7 +54,8 @@ Firecracker projects, or both.
 
 The first release comes with the following Rust modules:
 
-* aio: Safe wrapper over [`Linux AIO`](http://man7.org/linux/man-pages/man7/aio.7.html).
+* aio: Safe wrapper over
+  [`Linux AIO`](http://man7.org/linux/man-pages/man7/aio.7.html).
 
 * errno: Structures, helpers, and type definitions for working with
   [`errno`](http://man7.org/linux/man-pages/man3/errno.3.html).
@@ -79,8 +85,9 @@ The first release comes with the following Rust modules:
 * signal: Enums, traits and functions for working with
   [`signal`](http://man7.org/linux/man-pages/man7/signal.7.html).
 
-* sockctrl_msg: Wrapper for sending and receiving messages with file descriptors
-  on sockets that accept control messages (e.g. Unix domain sockets).
+* sockctrl_msg: Wrapper for sending and receiving messages with file
+  descriptors on sockets that accept control messages (e.g. Unix domain
+  sockets).
 
 * tempdir: Structure for handling temporary directories.
 
--- a/src/fam.rs
+++ b/src/fam.rs
@@ -18,6 +18,11 @@
 
 use std::mem::{self, size_of};
 
+#[cfg(feature = "with-serde")]
+use serde::de::{Deserialize, Deserializer};
+#[cfg(feature = "with-serde")]
+use serde::{Serialize, Serializer};
+
 /// Errors associated with the [`FamStructWrapper`](struct.FamStructWrapper.html) struct.
 #[derive(Clone, Debug)]
 pub enum Error {
@@ -417,6 +422,34 @@ impl<T: Default + FamStruct> From<Vec<T>> for FamStructWrapper<T> {
     }
 }
 
+#[cfg(feature = "with-serde")]
+impl<T: Default + FamStruct + Serialize> Serialize for FamStructWrapper<T>
+where
+    <T as FamStruct>::Entry: serde::Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.as_slice().serialize(serializer)
+    }
+}
+
+#[cfg(feature = "with-serde")]
+impl<'de, T: Default + FamStruct + Deserialize<'de>> Deserialize<'de> for FamStructWrapper<T>
+where
+    <T as FamStruct>::Entry: std::marker::Copy + serde::Deserialize<'de>,
+{
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+        <T as FamStruct>::Entry: std::marker::Copy + serde::Deserialize<'de>,
+    {
+        let entries_vec: Vec<<T as FamStruct>::Entry> = Vec::deserialize(deserializer)?;
+        Ok(FamStructWrapper::from_entries(entries_vec.as_slice()))
+    }
+}
+
 /// Generate `FamStruct` implementation for structs with flexible array member.
 #[macro_export]
 macro_rules! generate_fam_struct_impl {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Added conditionally compiled `serde` compatibility to `FamStructWrapper`.
```
acatangiu: nit: explicitly state the feature name here

src/fam.rs
```diff
@@ -417,6 +422,34 @@ impl<T: Default + FamStruct> From<Vec<T>> for FamStructWrapper<T> {
...
+#[cfg(feature = "with-serde")]
+impl<T: Default + FamStruct + Serialize> Serialize for FamStructWrapper<T>
+where
+    <T as FamStruct>::Entry: serde::Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.as_slice().serialize(serializer)
```
serban300: Here we lose any information other than the flexible array member and its length. For example for `MockFamStruct` we loose the value of the `padding` field. Is that acceptable ?

aghecenco: The padding loss is a requirement the way I see it; there's no need to serialize and deserialize padding, and we should aim to deserialize as little as possible into the process' memory anyway for strengthening.
The `FamStruct` shouldn't hold anything other than the array and its length anyway.

sameo:
@aghecenco Could you please rebase this PR on top of master so that we don't push a merge point?

andreeaflorescu:
@sameo the merge commit will not be pushed to master. GitHub magic :D

sameo:
> @sameo the merge commit will not be pushed to master. GitHub magic :D

Because we will push the "rebase and merge" button when merging that PR? I still believe that it's cleaner to not add merge points as part of the PR and will e.g. show that this PR is a single commit one. But that's cosmetic mostly.

sameo:
@aghecenco Could you please rebase that one once #72 is merged?

src/fam.rs
aghecenco: Because the `FamStructWrapper` was designed to encapsulate an array, any other members should be irrelevant.

jiangliu: I haven't gotten the point yet. Seems it's possible to do
```
#[derive(Serialize, Deserialize)]
Struct Header {
     len: u32;
     msg_type: u32;
     pub entries: __IncompleteArrayField<u32>,
}

imp FamStruct for Header {
       type Entry = u32;
      .....
}

type Message = FamStructWrapper<Header>;
let data = [0x1u32, 0x2u32];
let msg = Message::from_entries(&data);
let hdr = msg.as_mut_fam_struct();
hdr.type = MSG_TYPE_XXX;

send_message(serde_json::to_string(&wrapper).unwrap());
```

According to current implementation, the Header.msg_type will get lost on the receiving side.
And we can't implement Serialize/Deserialize for Message because we already have:
```
impl<T: Default + FamStruct + Serialize> Serialize for FamStructWrapper<T>
where
    <T as FamStruct>::Entry: serde::Serialize,
{
```

aghecenco: Right. That's valid; this serialization works with the same simplifying assumption as the macro that generates `impl FamStruct`: https://github.com/rust-vmm/vmm-sys-util/blob/133524c0ad66ad66f340e1a4b6b71595a4a65898/src/fam.rs#L422
namely that the `struct` that implements `FamStruct` *doesn't have* any other relevant members other than `len`.

jiangliu: So my suggestion is to serialize/deserialize the header element too, and use #[serde(skip)] to explicitly mark fields in the FamStruct struct we want to skip.

jiangliu: > Right. That's valid; this serialization works with the same simplifying assumption as the macro that generates `impl FamStruct`:
> 
> https://github.com/rust-vmm/vmm-sys-util/blob/133524c0ad66ad66f340e1a4b6b71595a4a65898/src/fam.rs#L422
> 
> 
> namely that the `struct` that implements `FamStruct` _doesn't have_ any other relevant members other than `len`.

I'm not sure the `generate_fam_struct_impl` has a constraint that header only has a `len` field.
The generate_fam_struct_impl provides methods to access the `len` field, but it's legal for the header to have other fields.

aghecenco: My point was that it doesn't have other *relevant* fields, given that it represents an array. It could have any number of fields, sure, but then if it was a custom struct, it wouldn't need FAM wrappers and customized serialization. We'd just implement it as a vec. FAM structs solve a very niche problem, specific to the autogenerated bindings, and for all the intents and purposes of that niche problem, nothing in the autogenerated struct *matters* except the `len` and array members.

aghecenco: I understand what you're aiming at, but can't see how serializing the header would reconcile with deserializing the entries (well not without complicating the deserializer - unnecessarily for now). Specifically: the `FamStructWrapper` code is pretty much tailored to suit bindings, and only leverages the "header" (first member in the entries array, which doesn't represent a real entry) as a workaround to store information about the actual entries that follow. The "header" is completely reconstructed at deserialization, as the `FamStructWrapper` is initialized via [`from_entries`](https://github.com/rust-vmm/vmm-sys-util/blob/c71af57ed8924b4c0e15a0ef711e8bb89e742346/src/fam.rs#L200). If we were to serialize the header too, deserialization would look like this:
1. deserialize header, save it for later
1. deserialize `Vec` of `FamStruct::Entry`
1. create `FamStructWrapper` via `from_entries`. This generates a new header, which will hold the `len` of the `Entry` vector. All other info is lost.
1. overwrite generated header with the one deserialized earlier

This solution addresses your concern and keeps `FamStruct` flexible, at the expense of some added complexity in the deserializer. How does that sound?

jiangliu: Maybe we could do it better with following steps:)
On serialization side:
1.1) serialize Vec of FamStruct::Entry,
1.2) serialize the header element,
On desrialization side:
2.1) deserialize Vec of FamStruct::Entry,
2.2) deserialize the header element.
If we care about the len of the decoded Vec, we could cross-checking header_element.len equals size of the decoded Vec.
Thank:)

aghecenco: I think we're thinking of the same idea in different terms, I'll just come up with the code for it, easier to review that way :grinning:

aghecenco: Hey @jiangliu PTAL, I replaced the ser/deser derives with handcrafted logic to handle headers correctly. The unit test in https://github.com/rust-vmm/vmm-sys-util/pull/66/commits/805a2f30cfd0ed61b15bbdf1e6d8b51a3d406572 shows that additional values in the FAM header are included, with no ill effect on the entries.
I also found a bug in the `__IncompleteArrayField` ser/deserialization. The field needs to be  `serde(skip)`ped in order for `serde_json` to understand the string.

aghecenco:
@acatangiu @sameo @serban300 @jiangliu please re-review. I made some changes to address full header (de)serialization.

src/lib.rs
```diff
@@ -16,3 +16,8 @@ pub mod errno;
 pub mod fam;
 pub mod rand;
 pub mod tempfile;
+
+#[cfg(feature = "with-serde")]
```
jiangliu: How about moving the near to the top?

src/fam.rs
```diff
@@ -16,8 +16,17 @@
...
 
+#[cfg(feature = "with-serde")]
+use serde::de::{self, Deserialize, Deserializer, SeqAccess, Visitor};
+#[cfg(feature = "with-serde")]
+use std::fmt;
+#[cfg(feature = "with-serde")]
+use std::marker::PhantomData;
 use std::mem::{self, size_of};
 
+#[cfg(feature = "with-serde")]
```
jiangliu: How about moving line 27-28 near to line 20?

jiangliu:
> @acatangiu @sameo @serban300 @jiangliu please re-review. I made some changes to address full header (de)serialization.

Much more complex than expected, deserialization is always complex:)

aghecenco:
> Much more complex than expected, deserialization is always complex:)

Yeah, it turned out quite a bit more convoluted than before, just to get the header right and preserve the entries. Custom deserialization with serde is hard

Issue #58, opened 2019-12-04
the trait `std::error::Error` is not implemented for `vmm_sys_util::errno::Error`


haraldh:
SSIA :smile:

MrXinWang:
Trying to work on this :)

If I understand this subject correctly (Please do correct me if I am wrong :P), we only need to implement [`fn source(&self) -> Option<&(dyn Error + 'static)>`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) as there is an existing implementation of [`Display`](https://github.com/rust-vmm/vmm-sys-util/blob/664bb8df997b29c00edc8b32a90639e6158b68f6/src/errno.rs#L100) and [`fn backtrace(&self) -> Option<&Backtrace>`](https://doc.rust-lang.org/std/error/trait.Error.html#method.backtrace) is a nightly feature.

PR #67, opened 2019-12-30
Implement trait std::error::Error for vmm_sys_util::errno::Error

MrXinWang:
Referencing issue #58 , a basic implementation is proposed in this PR.

1. Implemented method `source` from `std::error::Error` for the custom error type `vmm_sys_util::errno::Error`.

2. Added unit test code for above implementation

Personally I think: Since the custom error type `pub struct Error(i32)` is just a wrapper of standard error, therefore there is no underlying error, hence we should always return `None` in the `source` method. Please do correct me if I am wrong. Thanks very much!

Signed-off-by: Henry Wang <<henry.wang@arm.com>>

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -9,6 +9,7 @@
 //! Structures, helpers, and type definitions for working with
 //! [`errno`](http://man7.org/linux/man-pages/man3/errno.3.html).
 
+use std::error::Error as StdError;
 use std::fmt::{Display, Formatter};
 use std::io;
 use std::result;
@@ -103,6 +104,12 @@ impl Display for Error {
     }
 }
 
+impl StdError for Error {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        None
+    }
+}
+
 impl From<io::Error> for Error {
     fn from(e: io::Error) -> Self {
         Error::new(e.raw_os_error().unwrap_or_default())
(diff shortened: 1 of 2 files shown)


src/errno.rs
```diff
@@ -135,6 +142,7 @@ mod tests {
 
         // Test that the inner value of `Error` corresponds to libc::EBADF.
         assert_eq!(last_err.errno(), libc::EBADF);
+        assert_eq!(last_err.source().is_none(), true);
```
aghecenco:
```suggestion
        assert!(last_err.source().is_none());
```

MrXinWang: Done. Thanks :)

MrXinWang:
@jiangliu Could you please take another look? Thanks!

aghecenco:
Hey @MrXinWang, I think you should squash the commits, as the 2nd change is minor. Good to merge after that :+1:

MrXinWang:
Hi @aghecenco ! Already squashed these two commits into a single one :) Thanks

src/errno.rs
```diff
@@ -103,6 +104,12 @@ impl Display for Error {
     }
 }
 
+impl StdError for Error {
```
haraldh: Huh? None is the trait default.. No need to implement that

fn source(&self) -> Option<&(dyn Error + 'static)> { None }

haraldh:
`impl StdError for Error {}` would have sufficed, I think

haraldh:
or even shorter and more clear to the reader without the `use ... as StdError`:
```rust
impl std::error::Error for Error {}
```

andreeaflorescu:
@haraldh since this commit is already merged, do you want to submit a fix in a PR?

src/errno.rs
MrXinWang: ah right it makes sense.....didn't thought of that as I only checked the [trait defination](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) in the offical doc. Thanks for pointing it out. 

I already saw your PR to fix it, thanks for that too :)

PR #68, opened 2020-01-06
signal: Block restricted signals in signal handler

KarthikNedunchezhiyan:
Blocked restricted signals from signal handling as like mentioned in [http://man7.org/linux/man-pages/man2/sigaction.2.html](url)

Signed-off-by: karthik nedunchezhiyan <karthik1705.n@gmail.com>

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -156,6 +156,13 @@ pub fn validate_signal_num(num: c_int) -> errno::Result<()> {
 pub fn register_signal_handler(num: c_int, handler: SignalHandler) -> errno::Result<()> {
     validate_signal_num(num)?;
 
+    // signum specifies the signal and can be any valid signal except
+    // SIGKILL and SIGSTOP.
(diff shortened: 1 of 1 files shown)


src/signal.rs
```diff
@@ -156,6 +156,13 @@ pub fn validate_signal_num(num: c_int) -> errno::Result<()> {
 pub fn register_signal_handler(num: c_int, handler: SignalHandler) -> errno::Result<()> {
     validate_signal_num(num)?;
 
+    // signum specifies the signal and can be any valid signal except
```
andreeaflorescu: Should this check be in validate_signal_num as well?

KarthikNedunchezhiyan: my opinion is not needed because _validate_signal_num_ is generic util and this restriction only applies specific to _register_signal_handler_ alone 🙂

andreeaflorescu: Ok.

KarthikNedunchezhiyan: @andreeaflorescu I have covered unit test 🙂

PR #69, opened 2020-01-09
src/errno.rs: remove bogus impl of std::error::Error::source()

haraldh:
resolves https://github.com/rust-vmm/vmm-sys-util/pull/67#issuecomment-572535329
and
https://github.com/rust-vmm/vmm-sys-util/issues/58

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -9,7 +9,6 @@
 //! Structures, helpers, and type definitions for working with
 //! [`errno`](http://man7.org/linux/man-pages/man3/errno.3.html).
 
-use std::error::Error as StdError;
 use std::fmt::{Display, Formatter};
 use std::io;
 use std::result;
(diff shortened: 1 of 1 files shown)


haraldh:
> Please sign your commit.
> 
> Can you also fix the style as specified here: https://buildkite.com/rust-vmm/vmm-sys-util-ci/builds/283#9f931888-539e-4948-a531-df01d0ff6af4
> 
> In the commit message, we are trying to adhere to the [50/70 rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule) where possible. Can you rephrase the commit summary to be shorter? I don't think we need all details in the summary, so you could probably write something like "Simplified implementation of Error". You could also write in the commit body something like: "Fixes #67" and when the PR is merged, the issue will be automatically closed.

fixed..

Issue #64, opened 2019-12-11
TempFile and TempDir support for file operations


iulianbarbu:
Two options worth considering:

1. Add support for temporary file operations, like read, write, seek, but also support for file permissions (open TempFile with specific file permissions). Some progress was done in this direction: https://github.com/rust-vmm/vmm-sys-util/pull/60.

2. When creating a temporary file/directory, we should unlink it immediately, then return a File that represent the handle for that unliked file/directory, that we're gonna operate with, and leave altogether the TempFile/TempDir structs as builders for these temporary files. This would simplify a lot the implementation, and wouldn't need any support for file operations and permissions.

acatangiu:
I don't agree with option 2. There are cases where one is interested in using the temp file path, unlinking the file on creation makes us lose the the property of "having temporary file `F` at the temporary path `P`".

Let's look at the multiple possible use-cases and make the right decision. We might find that solving all use-cases using a single object type is not possible.

From a bird's eye view option 1 should satisfy all requirements, **if** we implement all relevant `io` traits for the `TempFile` wrapper. On top of those, `TempFile` should also offer a `fn into_file(self) -> File` method that provides the owned `File` (the method consumes `self` so it also unlinks the file).

aghecenco:
What's the problem we want to solve?
* there are use cases where we need `File` operations
* there are use cases when we just need a random tmp path

I would leave `TempFile` stripped bare of functionality and immediately unlinking but exposing both an
* `into_file` which returns a `File` that we can operate on normally
* `into_path` which returns a `str` that we can use to manually control the file's lifecycle

both consuming the `TempFile`.

WDYT?

acatangiu:
I would like to give the user some control on when the file is unlinked from the file system. That’s why I don’t like any of the options that immediately unlink.

Let’s say for example you want to use a TempFile and have a reader and a writer in your code both of which requiring owning the file. With normal files you can achieve this by simply opening the same file twice each open giving you an owned File object. If we immediately unlink the TempFile, we lose this expected file behaviour.

The more I think about it, the more I support the idea that TempFile should be an “invisible” wrapper over File (meaning it should implement io traits). TempFile specific behaviour is creating the file in TEMPDIR and unlinking the file on Drop.

alexandruag:
[...]

PR #70, opened 2020-01-15
[TempFile] add into_file() method

serban300:
This method consumes the TempFile returning the wrapped file.
It also removes the file from the system. The file descriptor remains
opened and it can be used until the returned file is dropped.

Adding this method enables us to [drop the dependency on the tempfile crate
in favor of vmm-sys-util inside vm-memory](https://github.com/rust-vmm/vm-memory/issues/23). This helps us to remove about
17 sub-dependencies.

### DIFF
--- a/src/tempfile.rs
+++ b/src/tempfile.rs
@@ -36,7 +36,7 @@ use crate::rand::rand_alphanumerics;
 /// The file will be maintained for the lifetime of the `TempFile` object.
 pub struct TempFile {
     path: PathBuf,
-    file: File,
+    file: Option<File>,
 }
 
 impl TempFile {
@@ -64,7 +64,7 @@ impl TempFile {
 
         Ok(TempFile {
             path: file_path_buf,
-            file,
+            file: Some(file),
         })
     }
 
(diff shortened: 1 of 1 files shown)


src/tempfile.rs
```diff
@@ -121,7 +121,14 @@ impl TempFile {
 
     /// Returns a reference to the File
     pub fn as_file(&self) -> &File {
-        &self.file
+        // It's safe to unwrap because `file` can be `None` only after calling `into_file`
+        // which consumes this object.
+        self.file.as_ref().unwrap()
+    }
+
+    /// Consumes the TempFile, returning the wrapped file.
```
bonzini: I would add to the doc comment that this also removes the file!

serban300: Done.

serban300:
> Please document `into_file()` so that users know that the `TemFile.file` will be removed after calling that function.
> Also, could you please also explain in a little more details, in the commit message, how this helps with reducing the vm-memory dependencies reduction? By just reading the commit message, it's not clear that this will allow the vm-memory crate to completely drop the `tmpfile` crate dependency in favor of vmm-sys-utils (which should bring less sub-dependencies...)

@sameo I addressed the comments. Please take another look !

andreeaflorescu:
@sameo and @serban300 I don't think we need to specify in the commit message that this enables us to drop the tempfile dependency in vm-memory.

I would rather have the commit message specify that it fixes this problem: https://github.com/rust-vmm/vmm-sys-util/issues/64. Isn't that the reason for this PR?

sameo:
> @sameo and @serban300 I don't think we need to specify in the commit message that this enables us to drop the tempfile dependency in vm-memory.
> 
> I would rather have the commit message specify that it fixes this problem: #64. Isn't that the reason for this PR?

It does fix #64 by implementing option #2. Having that stated in the commit message would indeed look better.

serban300:
@andreeaflorescu @sameo I changed the commit message. Please take a look.

PR #71, opened 2020-01-16
Make TempFile work on Windows

serban300:
Reimplement new_with_prefix() using rand::rand_alphanumerics() in order
to make it work on Windows.

Small fixes for the unit tests in order to run on Windows.

### DIFF
--- a/src/tempfile.rs
+++ b/src/tempfile.rs
@@ -23,16 +23,13 @@
 //! f.sync_all().unwrap();
 
 use std::env::temp_dir;
-use std::ffi::{CString, OsStr};
+use std::ffi::OsStr;
 use std::fs;
-use std::fs::File;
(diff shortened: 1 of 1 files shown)


src/tempfile.rs
```diff
@@ -142,16 +130,16 @@ mod tests {
             (to_check >= lower) && (to_check <= upper)
         }
 
-        let tempname = "/tmp/asdf";
-        let t = TempFile::new_with_prefix(tempname).unwrap();
-        assert_eq!(tempname, "/tmp/asdf");
+        let mut prefix = temp_dir();
+        prefix.push("asdf");
```
andreeaflorescu: Why do you need this push?

serban300: Just wanted to keep the previous behavior, namely to use the prefix `/tmp/asdf` (on Linux).

PR #72, opened 2020-01-20
enable Windows build

serban300:


src/lib.rs
```diff
@@ -7,22 +7,34 @@
...
+#[cfg(unix)]
 pub mod signal;
+#[cfg(unix)]
 pub mod sock_ctrl_msg;
+#[cfg(unix)]
 pub mod tempdir;
 pub mod tempfile;
+#[cfg(unix)]
 pub mod terminal;
+#[cfg(unix)]
```
sameo: Could we group all the `#[cfg(unix)]` together in the same part of the `lib.rs` file, to make it clearer which modules are unix specific?

andreeaflorescu:
@serban300 How about we first merge #70 and then enable the Windows CI?

serban300:
> @serban300 How about we first merge #70 and then enable the Windows CI?

They are orthogonal. We can prioritize them as you wish, but we need both in order to remove the `tempfile` dependency in `vm-memory`.

src/lib.rs
serban300: Done

andreeaflorescu:
@serban300 if we first enable the Windows build, then into_file PR will not pass the tests any more because it will fail on Windows. That's why I was thinking about doing them in the other order.

serban300:
> @serban300 if we first enable the Windows build, then into_file PR will not pass the tests any more because it will fail on Windows. That's why I was thinking about doing them in the other order.

Ok. Sounds good to me.

sameo:
@serban300 Please rebase and I'll approve it.

src/lib.rs
andreeaflorescu: I still think it doesn't look very nice as it is now. I had another idea in mind.

My proposal would be to create a new `unix` module and move all the unix specific utilities in that module. In `src/lib.rs` we would just export everything from unix only when on an unix platform.

In src/lib.rs we would just have:
```rust
#[cfg(unix)]
mod unix;

#[cfg(unix)]
pub use unix::*;

pub mod fam;
pub mod file_traits;
pub mod rand;
pub mod tempfile;
pub mod write_zeroes;
```

And in `src/unix/mod.rs`:
```rust
#[macro_use]
pub mod ioctl;
pub mod aio;
pub mod errno;
pub mod eventfd;
pub mod fallocate;
pub mod poll;
pub mod seek_hole;
pub mod signal;
pub mod sock_ctrl_msg;
pub mod tempdir;
pub mod terminal;
pub mod timerfd;
```

Nothing else needs to be changed and the vmm-sys-util exports the modules in the same way as before.

andreeaflorescu:
The README also needs to be updated to state that we also support Windows for a few of the utilities.

src/lib.rs
sameo: > My proposal would be to create a new unix module and move all the unix specific utilities in that module

I thought about this but assumed this would force caller path inclusion changes. If there's a way to avoid that and move everything under a unix specific module, then I'm all for it.

andreeaflorescu: This line avoids the path inclusion: `pub use unix::*;`

jiangliu: It's really a little ugly, agree to move them into a dedicate xxx_unix.rs file.

andreeaflorescu:
@serban300 please also update the readme.

serban300:
@andreeaflorescu @jiangliu @sameo I addressed all your comments. Please take another look !

README.md
```diff
@@ -14,6 +14,7 @@ with files, event file descriptors, ioctls and others.
 
 **Operating Systems**:
 - Linux
+- Windows (partial support)
```
andreeaflorescu: We also need to update this phrase: "The crate implements safe wrappers around common Linux utilities for working
with files, event file descriptors, ioctls and others."

serban300: Done

PR #74, opened 2020-01-30
signal: add functionality that waits for a signal

aghecenco:
Added a function that suspends execution until a specified signal (or for any signal at all) becomes pending, either indefinitely (with [`sigwaitinfo`](https://linux.die.net/man/2/sigtimedwait)) or with a timeout (with [`sigtimedwait`](https://linux.die.net/man/2/sigtimedwait)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+* Added the `signal::wait_for_signals` function, which suspends execution until
+  either a specified signal or any signal becomes pending, either indefinitely
+  or with a timeout.
+
 # v0.5.0
 
 * Added conditionally compiled `serde` compatibility to `FamStructWrapper`,
--- a/src/unix/signal.rs
+++ b/src/unix/signal.rs
@@ -11,14 +11,15 @@
 
 use libc::{
     c_int, c_void, pthread_kill, pthread_sigmask, pthread_t, sigaction, sigaddset, sigemptyset,
-    sigfillset, siginfo_t, sigismember, sigpending, sigset_t, sigtimedwait, timespec, EAGAIN,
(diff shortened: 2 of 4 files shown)


aghecenco:
@jiangliu lost your approval with the force-push, can you take another look?

CHANGELOG.md
```diff
@@ -85,9 +88,15 @@ The first release comes with the following Rust modules:
 * signal: Enums, traits and functions for working with
   [`signal`](http://man7.org/linux/man-pages/man7/signal.7.html).
 
+<<<<<<< HEAD
```
bonzini: You have some leftover conflict markings.

aghecenco:
@bonzini @jiangliu fixed & rebased, can you please take another look?

src/unix/signal.rs
```diff
@@ -402,6 +413,77 @@ pub fn clear_signal(num: c_int) -> SignalResult<()> {
...
+        let duration = timespec {
+            tv_sec: nanos / 1_000_000_000,
+            tv_nsec: nanos % 1_000_000_000,
+        };
+        let ret = unsafe { sigtimedwait(&sigset, null_mut(), &duration) };
+        if ret < 0 {
+            let e = errno::Error::last();
+            match e.errno() {
+                // `EAGAIN` means no signal became pending in the duration specified.
+                EAGAIN => (),
```
jiangliu, line 471: Do we need to return the TimedOut condition to caller?

aghecenco: `EAGAIN` is not an error condition, it's an indicator that no signal stopped the wait. Any other error (according to the man page, it can be either `EINVAL` or `EINTR`) will be passed to the user 2 lines below.

aghecenco:
@bonzini @jiangliu can you please take another look?

PR #79, opened 2020-03-12
Add epoll wrapper

lauralt:
Added small wrapper over libc::epoll that can be used for the basic operations regarding the epoll
API.

### DIFF
--- a/src/unix/epoll.rs
+++ b/src/unix/epoll.rs
@@ -1,0 +1,520 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Safe wrappers over the
+//! [`epoll`](http://man7.org/linux/man-pages/man7/epoll.7.html) API.
+
+use std::io;
+use std::ops::{Deref, Drop};
+use std::os::unix::io::{AsRawFd, RawFd};
+
+use libc::{
+    epoll_create1, epoll_ctl, epoll_event, epoll_wait, EPOLLERR, EPOLLET, EPOLLEXCLUSIVE, EPOLLHUP,
+    EPOLLIN, EPOLLONESHOT, EPOLLOUT, EPOLLPRI, EPOLLRDHUP, EPOLLWAKEUP, EPOLL_CLOEXEC,
+    EPOLL_CTL_ADD, EPOLL_CTL_DEL, EPOLL_CTL_MOD,
+};
+
+use syscall::SyscallReturnCode;
+
+/// Wrapper over `EPOLL_CTL_*` operations that can be performed on a file descriptor.
+#[repr(i32)]
+pub enum ControlOperation {
+    /// Add a file descriptor to the interest list.
+    Add = EPOLL_CTL_ADD,
+    /// Change the settings associated with a file descriptor that is
+    /// already in the interest list.
+    Modify = EPOLL_CTL_MOD,
+    /// Remove a file descriptor from the interest list.
+    Delete = EPOLL_CTL_DEL,
+}
+
+bitflags! {
+    /// The type of events we can monitor a file descriptor for.
+    pub struct EventSet: u32 {
+        /// The associated file descriptor is available for read operations.
+        const IN = EPOLLIN as u32;
+        /// The associated file descriptor is available for write operations.
+        const OUT = EPOLLOUT as u32;
+        /// Error condition happened on the associated file descriptor.
+        const ERROR = EPOLLERR as u32;
+        /// This can be used to detect peer shutdown when using Edge Triggered monitoring.
+        const READ_HANG_UP = EPOLLRDHUP as u32;
+        /// Sets the Edge Triggered behavior for the associated file descriptor.
+        /// The default behavior is Level Triggered.
+        const EDGE_TRIGGERED = EPOLLET as u32;
+        /// Hang up happened on the associated file descriptor. Note that `epoll_wait`
+        /// will always wait for this event and it is not necessary to set it in events.
+        const HANG_UP = EPOLLHUP as u32;
+        /// There is an exceptional condition on that file descriptor. It is mostly used to
+        /// set high priority for some data.
+        const PRIORITY = EPOLLPRI as u32;
+        /// The event is considered as being "processed" from the time when it is returned
+        /// by a call to `epoll_wait` until the next call to `epoll_wait` on the same
+        /// epoll file descriptor, the closure of that file descriptor, the removal of the
+        /// event file descriptor via EPOLL_CTL_DEL, or the clearing of EPOLLWAKEUP
+        /// for the event file descriptor via EPOLL_CTL_MOD.
+        const WAKE_UP = EPOLLWAKEUP as u32;
+        /// Sets the one-shot behavior for the associated file descriptor.
+        const ONE_SHOT = EPOLLONESHOT as u32;
+        /// Sets an exclusive wake up mode for the epoll file descriptor that is being
+        /// attached to the associated file descriptor.
+        /// When a wake up event occurs and multiple epoll file descriptors are attached to
+        /// the same target file using this mode, one or more of the epoll file descriptors
+        /// will receive an event with `epoll_wait`. The default here is for all those file
+        /// descriptors to receive an event.
+        const EXCLUSIVE = EPOLLEXCLUSIVE as u32;
+    }
+}
+
+/// Wrapper over
+/// ['libc::epoll_event'](https://doc.rust-lang.org/1.8.0/libc/struct.epoll_event.html).
+// We are using `transparent` here to be super sure that this struct and its fields
+// have the same alignment as those from the `epoll_event` struct from C.
+#[repr(transparent)]
+#[derive(Clone)]
+pub struct EpollEvent(epoll_event);
+
+impl Deref for EpollEvent {
+    type Target = epoll_event;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Default for EpollEvent {
+    fn default() -> Self {
+        EpollEvent(epoll_event {
+            events: 0u32,
+            u64: 0u64,
+        })
+    }
+}
+
+impl EpollEvent {
+    /// Create a new epoll_event instance.
+    ///
+    /// # Arguments
+    ///
+    /// `events` - contains an event mask.
+    /// `data` - a user data variable. `data` field can be a fd on which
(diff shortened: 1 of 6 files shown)


src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+use std::os::unix::io::{AsRawFd, RawFd};
+
+use libc::{
+    epoll_create1, epoll_ctl, epoll_event, epoll_wait, EPOLLERR, EPOLLET, EPOLLEXCLUSIVE, EPOLLHUP,
+    EPOLLIN, EPOLLONESHOT, EPOLLOUT, EPOLLPRI, EPOLLRDHUP, EPOLLWAKEUP, EPOLL_CLOEXEC,
+    EPOLL_CTL_ADD, EPOLL_CTL_DEL, EPOLL_CTL_MOD,
+};
+
+/// Wrapper to interpret syscall exit codes and provide a rustacean `io::Result`
+pub struct SyscallReturnCode(pub c_int);
```
aghecenco: This shouldn't be in the `epoll` mod but separate. It operates with any syscall.

src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+            Ok(self.0)
+        }
+    }
+    /// Returns the last OS error if value is -1 or Ok(()) otherwise.
+    pub fn into_empty_result(self) -> std::io::Result<()> {
+        self.into_result().map(|_| ())
+    }
+}
+
+/// Wrapper over EPOLL_CTL_* operations that can be performed on a file descriptor.
```
aghecenco:
```suggestion
/// Wrapper over `EPOLL_CTL_*` operations that can be performed on a file descriptor.
```

src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+        /// attached to the associated file descriptor.
+        /// When a wake up event occurs and multiple epoll file descriptors are attached to
+        /// the same target file using this mode, one or more of the epoll file descriptors
+        /// will receive an event with `epoll_wait`. The default here is for all those file
+        /// descriptors to receive an event.
+        const EXCLUSIVE = EPOLLEXCLUSIVE as u32;
+    }
+}
+
+/// Wrapper over 'libc::epoll_event'.
```
aghecenco:
```suggestion
/// Wrapper over ['libc::epoll_event'](https://doc.rust-lang.org/latest/libc/struct.epoll_event.html).
```

src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+    fn default() -> Self {
+        EpollEvent(epoll_event {
+            events: 0u32,
+            u64: 0u64,
+        })
+    }
+}
+
+impl EpollEvent {
+    /// Create a new epoll_event instance with the following fields: `events`, which contains
```
aghecenco: Please restructure this to respect the doccomments format:
```rust
/// # Arguments
///
/// * `events` - ....
```

src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// an event mask and `data` which represents a user data variable. `data` field can be
+    /// a fd on which we want to monitor the events specified by `events`.
+    pub fn new(events: EventSet, data: u64) -> Self {
+        EpollEvent(epoll_event {
+            events: events.bits(),
+            u64: data,
+        })
+    }
+
+    /// Returns the `events` from `libc::epoll_event`.
```
aghecenco: Wherever there's a link-able reference to another crate, please add the link so that the generated documentation is easily navigable. See https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html.

src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+    ///
+    /// This can be used for adding, modifying or removing a file descriptor in the
+    /// interest list of the epoll instance.
+    ///
+    /// # Arguments
+    ///
+    /// * `operation` refers to the action to be performed on the file descriptor.
+    /// * `fd` is the file descriptor on which we want to perform `operation`.
+    /// * `event` refers to the `epoll_event` instance that is linked to `fd`.
+    pub fn ctl(
```
aghecenco: Can you add `/// # Examples` sections to the public facing APIs too?

src/unix/epoll.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+}
+
+impl AsRawFd for Epoll {
+    fn as_raw_fd(&self) -> RawFd {
+        self.epoll_fd
+    }
+}
+
+impl std::ops::Drop for Epoll {
```
aghecenco: nit: `use std::ops::Drop` and skip the `std::ops` to be consistent with the other implementations

src/unix/epoll.rs
lauralt: I moved it to a new file, `syscall.rs`, as I didn't find a suitable place in the existing files. What do you think?

src/unix/epoll.rs
lauralt: Done.

src/unix/epoll.rs
lauralt: Done.

src/unix/epoll.rs
lauralt: Done.

src/unix/epoll.rs
lauralt: :+1:

src/unix/epoll.rs
lauralt: Done.

src/unix/epoll.rs
lauralt: :+1:

andreeaflorescu:
@sameo @jiangliu can you also take a quick look over this PR? The plan would be to eventually get rid of `poll.rs` module and use this safe wrapper instead. I assume right now this is not possible because probably you are using `EpollContext`. But if we switch to the vmm-epoll (see PR: https://github.com/rust-vmm/vmm-epoll/pull/2) then `EpollContext` and all the other wrappers in `poll.rs` would not be needed any more. Alternatively, we could keep the existing functionality in `poll.rs` and just remove the dependency on the epoll crate. WDYT?

src/unix/epoll.rs
```diff
@@ -0,0 +1,535 @@
...
+/// Wrapper over epoll functionality.
+#[derive(Debug)]
+pub struct Epoll {
+    epoll_fd: RawFd,
+}
+
+impl Epoll {
+    /// Create a new epoll file descriptor.
+    ///
+    /// # Examples
```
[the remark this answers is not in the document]
sameo: I agree here.

sameo:
> @sameo @jiangliu can you also take a quick look over this PR? The plan would be to eventually get rid of `poll.rs` module and use this safe wrapper instead. I assume right now this is not possible because probably you are using `EpollContext`. But if we switch to the vmm-epoll (see PR: [rust-vmm/vmm-epoll#2](https://github.com/rust-vmm/vmm-epoll/pull/2)) then `EpollContext` and all the other wrappers in `poll.rs` would not be needed any more. Alternatively, we could keep the existing functionality in `poll.rs` and just remove the dependency on the epoll crate. WDYT?

@andreeaflorescu I'm ok with this PR. Is the plan to have `vmm-epoll` use that module and basically move https://github.com/rust-vmm/vmm-epoll/pull/2/files#diff-1a41e4dcb6cec5befecca71500e4d036 here? If so, could we enhance this PR with the missing bits from the `vmm-epoll/src/epoll.rs` file?
Although we would not be consuming this module directly from Cloud-Hypervisor, this makes sense to me.

sameo:
> > @sameo @jiangliu can you also take a quick look over this PR? The plan would be to eventually get rid of `poll.rs` module and use this safe wrapper instead. I assume right now this is not possible because probably you are using `EpollContext`. But if we switch to the vmm-epoll (see PR: [rust-vmm/vmm-epoll#2](https://github.com/rust-vmm/vmm-epoll/pull/2)) then `EpollContext` and all the other wrappers in `poll.rs` would not be needed any more. Alternatively, we could keep the existing functionality in `poll.rs` and just remove the dependency on the epoll crate. WDYT?
> 
> @andreeaflorescu I'm ok with this PR. Is the plan to have `vmm-epoll` use that module and basically move https://github.com/rust-vmm/vmm-epoll/pull/2/files#diff-1a41e4dcb6cec5befecca71500e4d036 here? If so, could we enhance this PR with the missing bits from the `vmm-epoll/src/epoll.rs` file?

Nevermind, it is complete. Once we remove the chatty `New()` example, I'll add my approval.

andreeaflorescu:
> > @sameo @jiangliu can you also take a quick look over this PR? The plan would be to eventually get rid of `poll.rs` module and use this safe wrapper instead. I assume right now this is not possible because probably you are using `EpollContext`. But if we switch to the vmm-epoll (see PR: [rust-vmm/vmm-epoll#2](https://github.com/rust-vmm/vmm-epoll/pull/2)) then `EpollContext` and all the other wrappers in `poll.rs` would not be needed any more. Alternatively, we could keep the existing functionality in `poll.rs` and just remove the dependency on the epoll crate. WDYT?
> 
> @andreeaflorescu I'm ok with this PR. Is the plan to have `vmm-epoll` use that module and basically move https://github.com/rust-vmm/vmm-epoll/pull/2/files#diff-1a41e4dcb6cec5befecca71500e4d036 here? If so, could we enhance this PR with the missing bits from the `vmm-epoll/src/epoll.rs` file?
> Although we would not be consuming this module directly from Cloud-Hypervisor, this PR LGTM.

@sameo yap, the plan is to use it from vmm-epoll. If CH still wants to use poll.rs, I think we should improve it and remove the dependency on the external crate `epoll`. If long term you also plan to switch to vmm-epoll, then there is no point in investing that effort in `poll.rs` any more.

src/unix/epoll.rs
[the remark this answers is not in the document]
lauralt: I removed that example.

src/unix/epoll.rs
```diff
@@ -0,0 +1,520 @@
...
+    /// # Arguments
+    ///
+    /// `events` - contains an event mask.
+    /// `data` - a user data variable. `data` field can be a fd on which
+    ///          we want to monitor the events specified by `events`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// extern crate vmm_sys_util;
```
aghecenco: To be consistent with other crates, the `extern crate`s and `use`s should be hidden from the documentation. But I see that the whole `vmm-sys-util` crate has this problem. I'll open a separate issue, it's way out of scope for this PR.

PR #83, opened 2020-04-01
Added some epoll derives & updated rust-vmm-ci

### LINKED ISSUE #84: Replace mem::uninitialized with mem::MaybeUninit
With Rust 1.39 warning checks failed because in `src/unix/aio.rs` we use the deprecated `mem::uninitialized`. This should be replaced with the safe version `mem::MaybeUninit`.

andreeaflorescu:
Some derives are useful (and needed by vmm-epoll).

Updated rust-vmm-ci as we were using  pretty old version.

Also fixed #84 because the build would fail otherwise.

### DIFF
--- a/coverage_config_aarch64.json
+++ b/coverage_config_aarch64.json
@@ -1,0 +1,5 @@
+{
+  "coverage_score": 84.2,
+  "exclude_path": "",
+  "crate_features": ""
+}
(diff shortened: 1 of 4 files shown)


andreeaflorescu:
@jiangliu can you please take another look? I had to fix the unit tests and a tiny bug :D

PR #86, opened 2020-04-17
Use target_os=linux for conditional compilation

allisonrandal:
Currently, this crate uses target_family=unix for
conditional compilation of sections of code that only work
on Linux. This causes problems on other Unix family
operating systems, which fail to compile the Linux-specific
code. Changing the check to target_os=linux correctly
identifies where that code is expected to work, and allows
non-Linux Unixes to use vmm-sys-util and vm-memory.

Signed-off-by: Allison Randal <allison@lohutok.net>


src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! [rust-vmm](https://github.com/rust-vmm/community) components.
 
 #![deny(missing_docs)]
-#[cfg(unix)]
+#[cfg(linux)]
 #[macro_use]
```
rbradford, line 9: Does bitflags only work on Linux?

allisonrandal: It's only used by the Linux-specific code, so when you don't compile in the 'unix' directory, bitflags is unused (so Rust complains about an unused dependency).

src/lib.rs
```diff
@@ -14,9 +14,9 @@ extern crate serde;
 #[cfg(feature = "with-serde")]
 extern crate serde_derive;
 
-#[cfg(unix)]
+#[cfg(linux)]
 mod unix;
```
jiangliu, line 18: Should we  also rename the mod to linux?

allisonrandal: Yes, it would make sense for clarity. I didn't make that part of this pull request, because it would move a lot of files. But, it wouldn't be a disruptive change, since the 'unix' directory is flattened by the compile (for example, anything that depends on the Linux-specific src/unix/eventfd.rs uses "vmm_sys_util::eventfd" and not "vmm_sys_uti::unix::eventfd").

jiangliu: Some modules in src/unix are not linux specific, such as file_trait, signal, tempdir, sock_msg_ctl etc. So should we change src/unix/mod.rs to use "#[cfg(linux)]" for feature control?

allisonrandal: That approach is also possible, here's what it looks like:

```
diff --git a/src/unix/mod.rs b/src/unix/mod.rs
index 8bf3202..3c4e36a 100644
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1,15 +1,26 @@
+#[cfg(linux)]
 #[macro_use]
 pub mod ioctl;
+#[cfg(linux)]
 pub mod aio;
+#[cfg(linux)]
 pub mod epoll;
+#[cfg(linux)]
 pub mod eventfd;
+#[cfg(linux)]
 pub mod fallocate;
 pub mod file_traits;
+#[cfg(linux)]
 pub mod poll;
+#[cfg(linux)]
 pub mod seek_hole;
+#[cfg(linux)]
 pub mod signal;
+#[cfg(linux)]
 pub mod sock_ctrl_msg;
 pub mod tempdir;
 pub mod terminal;
+#[cfg(linux)]
 pub mod timerfd;
+#[cfg(linux)]
 pub mod write_zeroes;
```
So, only file_traits, tempdir, and terminal successfully compile on non-Linux Unixes.

If you want to take this approach, I suggest moving all Linux-specific modules to src/linux, and keeping the ones intended to be compatible with all Unixes in src/unix. That clean separation makes the conditional compilation more obvious and maintainable, with a simple [#cfg(unix)] in src/lib.rs to include src/unix (for all Unix family OSs), and a simple [#cfg(linux)] in src/lib.rs to include src/linux (only for Linux).

jiangliu: I'm ok to move to dedicated linux directory:) Need more feedbacks from other maintainers.
BTW, I expect sock_ctrl_msg should work on all unix too, maybe need to resolve some minor bugs.

allisonrandal: Nod, I'm happy to rework the patch to fit whatever approach the maintainers prefer.

The compile errors in sock_ctrl_msg are mostly type errors, specifically a usize type where the compiler expected either i32 or u32. It also uses unix/eventfd.rs, which also fails to compile on non-Linux, in that case because libc::eventfd doesn't exist, and neither do the constants libc::EFD_CLOEXEC, libc::EFD_NONBLOCK, and libc::EFD_SEMAPHORE.

So far, I haven't encountered use of sock_ctrl_msg in other rust-vmm crates, but eventfd is used extensively for device management in both Firecracker and Cloud Hypervisor, so it's likely I'll spend more time digging into that failure later.

allisonrandal: Looking into it a bit more, the OmniOS Illumos distro does have a Linux-style eventfd (added in the 2017 LTS release), but Rust's libc crate only conditionally compiles eventfd for Linux and Android.

andreeaflorescu: I would go forward as follows:
- add a linux module with the utilities that only build on Linux
- in the unix module only leave the utilities that build on Unix as well
- add GitHub issues for all Unix-compatible utilities that are now in the linux module, but could be "fixed" so that they properly work on Unix

This way @allisonrandal doesn't need to fix all the utilities. How does this sound?

jiangliu: Great, so we could move forward fast:)

allisonrandal: I've updated the pull request to follow the approach @andreeaflorescu suggested

src/lib.rs
jiangliu, line 9: Should we use "#[cfg(target_os = "linux")]" here?
Seems there's only shortcut for #[cfg("unix")] and no shortcut #[cfg("linux")], which then causes the huge change in code coverage.

allisonrandal: I wondered about that change in code coverage. The bitflags crate compiles fine on non-Linux, but because it isn't used, it gives a warning "unused `#[macro_use]` import".

Maybe a better solution is to add the shortcut for #[cfg("linux")]?

jiangliu: Sorry， I mean "#[cfg(linux)]" will always be false on linux system. We should replace all "#[cfg(linux)]" with "#[cfg(target_os = "linux")]:)

Cargo.toml
```diff
@@ -16,7 +16,7 @@ libc = ">=0.2.39"
 serde = { version = ">=1.0.27", optional = true }
 serde_derive = { version = ">=1.0.27", optional = true }
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(linux)'.dependencies]
```
andreeaflorescu, line 19: This should be `[target.'cfg(target_os = "linux")'.dependencies]`.

`cfg(linux)` is not defined (just unix is :))) ). I just noticed that because the coverage is increased (even though it shouldn't be). Every occurrence of `#[cfg(linux)]` should be replaced with `#[cfg(target_os = "linux")]`.

allisonrandal: Updated, thanks!

allisonrandal:
Apparently, ARM still has a change in test coverage, but only from 84.2 to 84.3, any thoughts on why?

jiangliu:
> Apparently, ARM still has a change in test coverage, but only from 84.2 to 84.3, any thoughts on why?

This should be ok, may be due to code changes.

andreeaflorescu:
@allisonrandal can you please update the arm coverage so we can merge this?

andreeaflorescu:
I was thinking about doing a release and it would be nice to also include this PR.

allisonrandal:
@andreeaflorescu Thanks for the reminder. I updated the arm coverage, and the checks all pass now.

PR #96, opened 2020-09-01
Add Illumos variants of some modules

allisonrandal:
This PR adds some variants of the Linux submodules, modified to work on Illumos. They depend on changes to Rust's libc released in version 0.2.74, and also depend on nightly Rust for the new target_os="illumos".

For now, this is a draft PR, because it still has some failing tests in src/illumos/seek_hole.rs, related to a few small but fundamental differences in the behavior of `lseek` between Linux and Illumos. I'd appreciate feedback on the approach, even though it's not quite ready to merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ license = "Apache-2.0 AND BSD-3-Clause"
 with-serde = ["serde", "serde_derive"]
 
 [dependencies]
-libc = ">=0.2.39"
+libc = ">=0.2.74"
 serde = { version = ">=1.0.27", optional = true }
 serde_derive = { version = ">=1.0.27", optional = true }
 
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(any(target_os = "linux", target_os = "illumos"))'.dependencies]
 bitflags = "1.0"
 
 [dev-dependencies]
--- a/src/illumos/epoll.rs
+++ b/src/illumos/epoll.rs
@@ -10,9 +10,9 @@ use std::os::unix::io::{AsRawFd, RawFd};
 
 use bitflags::bitflags;
 use libc::{
-    epoll_create1, epoll_ctl, epoll_event, epoll_wait, EPOLLERR, EPOLLET, EPOLLHUP, EPOLLIN,
-    EPOLLONESHOT, EPOLLOUT, EPOLLPRI, EPOLLRDHUP, EPOLLWAKEUP, EPOLL_CLOEXEC, EPOLL_CTL_ADD,
-    EPOLL_CTL_DEL, EPOLL_CTL_MOD,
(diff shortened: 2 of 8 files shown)


jiangliu:
> This PR adds some variants of the Linux submodules, modified to work on Illumos. They depend on changes to Rust's libc released in version 0.2.74, and also depend on nightly Rust for the new target_os="illumos".
> 
> For now, this is a draft PR, because it still has some failing tests in src/illumos/seek_hole.rs, related to a few small but fundamental differences in the behavior of `lseek` between Linux and Illumos. I'd appreciate feedback on the approach, even though it's not quite ready to merge.

Amazing, reminder me about the days with OpenSolaris:)

andreeaflorescu:
@allisonrandal is it possible to share the code between linux & illumos? It looks like the files in ilumos are almost identical to the ones in linux, so maybe it make sense to not have them both as they're now.

jiangliu:
> @allisonrandal is it possible to share the code between linux & illumos? It looks like the files in ilumos are almost identical to the ones in linux, so maybe it make sense to not have them both as they're now.

yes, how about organizing those duplicated code into "unix"?

allisonrandal:
@andreeaflorescu Yes, it should be possible to merge the linux and illumos variants. In fact, eventfd.rs is actually the same, since the changes to get it working went into Rust's libc, instead of requiring any changes in the vmm-sys-util code. The tricky ones are related to features that don't exist on Illumos:

- seek_hole.rs: different tests for different behavior around tmpfs and and sparse files
- signal.rs: drops the `clear_signal` and `is_pending` functions and related tests
- write_zeros.rs: drops the `PunchHole` trait and related implementation

If you all are comfortable carrying such large `cfg` blocks, I'll rework the patch to merge them.

@jiangliu Yes, the unix directory is probably the most logical place to put the merged files, though these won't work on Solaris, and I expect they won't work on the BSDs either. Illumos is similar to Linux in ways that other Unixes aren't. I'm guessing vmm-sys-util isn't currently being used by anyone on other Unixes, because they would have needed substantial patches to get it to compile. So, it's probably reasonable accept that `src/unix` is really only portable to the subset of Unixes where it's currently used and tested, and can always be split up later if someone comes along with a need for it.

Another question, are you all comfortable increasing the libc version requirement to 0.2.74 in Cargo.toml? The higher version is only necessary for Illumos.

allisonrandal:
Here's where the Illumos variants added in this PR are being used, in case anyone's curious (still a work in progress):
https://github.com/oxidecomputer/ch-oxidase

andreeaflorescu:
> Another question, are you all comfortable increasing the libc version requirement to 0.2.74 in Cargo.toml? The higher version is only necessary for Illumos.

Yes, that sounds good.

allisonrandal:
Dropping this draft pull request. The upstream Illumos devs are moving in a different direction for the Rust userspace for Bhyve.

PR #102, opened 2020-10-12
Add unit tests remove a directory that contains other directories #101

### LINKED ISSUE #101: tempdir: improve unit tests
The tests from `tempdir` are not covering the case when we remove a directory that contains other directories or files. We should add a test that checks this scenario too.

u5surf:
Signed-off-by: u5surf <u5.horie@gmail.com>

Fixes #101 

Adding unit tests which remove a directory that contains other directories

### DIFF
--- a/src/unix/tempdir.rs
+++ b/src/unix/tempdir.rs
@@ -160,12 +160,25 @@ mod tests {
 
     #[test]
     fn test_remove_dir() {
+        use crate::tempfile::TempFile;
         let t = TempDir::new().unwrap();
         let path = t.as_path().to_owned();
         assert!(t.remove().is_ok());
         // Calling remove twice returns error.
         assert!(t.remove().is_err());
         assert!(!path.exists());
+
+        let t = TempDir::new().unwrap();
+        let mut file = TempFile::new_in(t.as_path()).unwrap();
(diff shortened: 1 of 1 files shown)


src/unix/tempdir.rs
```diff
@@ -166,6 +166,14 @@ mod tests {
         // Calling remove twice returns error.
         assert!(t.remove().is_err());
         assert!(!path.exists());
+
+        let t = TempDir::new().unwrap();
+        let t2 = TempDir::new_in(t.as_path()).unwrap();
```
lauralt, line 171: I would also add a file in `t` dir:
```rust
let mut file = TempFile::new_in(t.as_path()).unwrap();
```
and do the same checks as with `t2`.

lauralt (commented):
Hi, @u5surf, thanks for your contribution!
The commit format test is failing, can you please update the title to something shorter, e.g.:

```
tempdir: improve unit tests

test_remove_dir() is now checking that removing a directory that
contains other directories and files is removing those too.
Fixes: #101.
```
?
See [reference](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule).

jiangliu (approved):

lauralt:
Looks good to me.

alexandruag (approved):

Issue #109, opened 2020-11-17
use FALLOC_FL_ZERO_RANGE for write_zeroes


lauralt:
While working on the virtio block discard/write_zeroes implementation, I noticed that `write_zeroes()` is using `fallocate` with FALLOC_FL_PUNCH_HOLE mode instead of FALLOC_FL_ZERO_RANGE. From [the virtio specification
](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-2500006) it makes sense to use FALLOC_FL_PUNCH_HOLE for write zeroes request only if `unmap` bit is also set, otherwise the zeroed region shouldn't be deallocated.

I think it would make sense to either update `write_zeroes()` with FALLOC_FL_ZERO_RANGE or add an extra parameter to that function (`mode` for example) which can be any `FallocateMode` variant.

Moreover, we can completely switch to [the latest crosvm `write_zeroes`](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/refs/heads/master/sys_util/src/write_zeroes.rs) which is using FALLOC_FL_ZERO_RANGE for `File` and adds a new trait `WriteZeroesAt` which lets you give the offset in file instead of using the current cursor. It also allows doing custom implementations for write_zeroes() (which might be useful for example for qcow file wrappers, example [here](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/refs/heads/master/disk/src/qcow/mod.rs#1714)).

Any thoughts / suggestions?

jiangliu:
The latest implementation from crosvm is preferred. Then the virtio-blk driver chooses to call punch_hole() or write_zeroes_at() according to the `unmap` flag for VIRTIO_BLK_T_WRITE_ZEROES requests.

lauralt:
That's how I was seeing it too, thanks! I'll open a PR with the switch.

PR #110, opened 2020-11-19
use FALLOC_FL_ZERO_RANGE for write_zeroes

lauralt:
#109 
More details in the commit message.

### DIFF
--- a/src/linux/write_zeroes.rs
+++ b/src/linux/write_zeroes.rs
@@ -1,3 +1,5 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
 // Copyright 2019 Intel Corporation. All Rights Reserved.
 //
 // Copyright 2018 The Chromium OS Authors. All rights reserved.
@@ -8,7 +10,8 @@
 
 use std::cmp::min;
 use std::fs::File;
-use std::io::{Error, Result, Seek, SeekFrom, Write};
+use std::io::{Error, ErrorKind, Result, Seek, SeekFrom};
+use std::os::unix::fs::FileExt;
 
 use crate::fallocate::{fallocate, FallocateMode};
 
@@ -19,7 +22,7 @@ pub trait PunchHole {
     /// # Arguments
     ///
     /// * `offset`: offset of the file where to replace with a hole.
-    /// * `length`: the bytes of a hole to replace with.
+    /// * `length`: the number of bytes of the hole to replace with.
     fn punch_hole(&mut self, offset: u64, length: u64) -> Result<()>;
 }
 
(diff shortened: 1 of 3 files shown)


src/linux/write_zeroes.rs
```diff
@@ -32,150 +35,276 @@ impl PunchHole for File {
...
-    /// Write `length` bytes of zeroes to the stream, returning how many
-    /// bytes were written.
+    /// Write up to `length` bytes of zeroes to the stream, returning how many bytes were written.
     fn write_zeroes(&mut self, length: usize) -> Result<usize>;
+
+    /// Write zeroes to the stream until `length` bytes have been written.
+    ///
+    /// This method will continuously call `write_zeroes` until the requested `length` is satisfied
+    /// or an error is encountered.
+    fn write_zeroes_all(&mut self, mut length: usize) -> Result<()> {
```
andreeaflorescu, line 45: This function name sounds a bit weird in English. How about we name it `write_all_zeroes`? This way it also is closer to the `write_all` nomenclature that seem like we want to follow here. Same goes for `write_zeroes_all_at`. I believe it sounds better as `write_all_zeroes_at`.

src/linux/write_zeroes.rs
```diff
@@ -32,150 +35,276 @@ impl PunchHole for File {
...
+/// A trait for writing zeroes to an arbitrary position in a file.
+pub trait WriteZeroesAt {
+    /// Write up to `length` bytes of zeroes starting at `offset`, returning how many bytes were
+    /// written.
+    fn write_zeroes_at(&mut self, offset: u64, length: usize) -> Result<usize>;
+
+    /// Write zeroes starting at `offset` until `length` bytes have been written.
+    ///
+    /// This method will continuously call `write_zeroes_at` until the requested `length` is
+    /// satisfied or an error is encountered.
```
andreeaflorescu, line 75: nit: "an unrecoverable error is encountered". We're handling the interrupt error. Same for `write_zero_all`.

src/linux/write_zeroes.rs
```diff
@@ -32,150 +35,276 @@ impl PunchHole for File {
...
+            }
         }
+        Ok(())
+    }
+}
 
-        // fall back to write()
+impl WriteZeroesAt for File {
+    fn write_zeroes_at(&mut self, offset: u64, length: usize) -> Result<usize> {
+        // Try to use fallocate() first.
```
andreeaflorescu, line 102: Let's also say that we are trying to use fallocate first because it is more efficient that writing zeroes.

src/linux/write_zeroes.rs
```diff
@@ -32,150 +35,276 @@ impl PunchHole for File {
...
-        // Bytes before the write should still be 0
-        for read in readback[0..1234].iter() {
+        // Bytes before the write should still be 0.
+        for read in &readback[0..1234] {
             assert_eq!(*read, 0);
         }
-        // Bytes that were just written should be 0x55
-        for read in readback[1234..(1234 + 5678)].iter() {
-            assert_eq!(*read, 0x55);
+        // Bytes that were just written should be 0x55.
```
andreeaflorescu, line 160: nit: we are not using a hardcoded value, so we need to update the comment as well.

src/linux/write_zeroes.rs
```diff
@@ -32,150 +35,276 @@ impl PunchHole for File {
...
+    fn test_small_write_zeroes() {
+        const NON_ZERO_VALUE: u8 = 0x55;
+
+        let mut f = TempFile::new().unwrap().into_file();
         f.set_len(16384).unwrap();
 
-        // Write buffer of non-zero bytes to offset 1234
-        let orig_data = [0x55u8; 5678];
+        // Write buffer of non-zero bytes to offset 1234.
+        let orig_data = [NON_ZERO_VALUE; 5678];
```
andreeaflorescu, line 148: Can we define a const for 5678 as well? It looks like it used multiple times.

src/linux/write_zeroes.rs
lauralt: Yep... I chose to keep the same name thinking that `write zeroes` is the action and would be easier to spot the difference between those function and the ones without `all`. Anyway, I updated with the english-friendly name.

src/linux/write_zeroes.rs
lauralt: done

src/linux/write_zeroes.rs
lauralt: good idea

src/linux/write_zeroes.rs
lauralt: done, also added for the file sizes in the other tests.

src/linux/write_zeroes.rs
lauralt: I left the numeric value here since it sounded more decent :)). Updated the comment.

src/linux/write_zeroes.rs
```diff
@@ -32,150 +35,282 @@ impl PunchHole for File {
...
-    /// Write zeroes to a stream.
-    ///
-    /// Write `length` bytes of zeroes to the stream, returning how many
-    /// bytes were written.
+    /// Write up to `length` bytes of zeroes to the stream, returning how many bytes were written.
     fn write_zeroes(&mut self, length: usize) -> Result<usize>;
+
+    /// Write zeroes to the stream until `length` bytes have been written.
+    ///
+    /// This method will continuously call `write_zeroes` until the requested `length` is satisfied
```
andreeaflorescu, line 43: We shouldn't leak implementation details in the documentation (i.e. that we call `write_zeroes`). We can say instead that "... will continuously write zeroes".

Same for `write_all_zeroes_at`.

andreeaflorescu (approved):

jiangliu (approved):

Issue #112, opened 2021-01-27
Unsound safe code could lead to UB in epoll.rs


Twister915:
In `src/linux/epoll.rs` the following function is defined:

```rust
pub fn wait(
    &self,
    max_events: usize,
    timeout: i32,
    events: &mut [EpollEvent],
) -> io::Result<usize> {
    // Safe because we give a valid epoll file descriptor and an array of epoll_event structures
    // that will be modified by the kernel to indicate information about the subset of file
    // descriptors in the interest list. We also check the return value.
    let events_count = SyscallReturnCode(unsafe {
        epoll_wait(
            self.epoll_fd,
            events.as_mut_ptr() as *mut epoll_event,
            max_events as i32,
            timeout,
        )
    })
    .into_result()? as usize;

    Ok(events_count)
}
```

It appears this is unsound, because there is no check that `max_events` is not greater than the size of the events slice. For example: if a call to `wait` is made `wait(128, -1, some_slice)` where `some_slice` is only large enough to hold 64 elements, then the kernel will write past the end of this slice leading to UB.

This function should therefore be marked `unsafe` or have an `assert!(events.len() >= max_events)` before the `unsafe` call to `epoll_wait`.

PR #113, opened 2021-01-27
Add length check to Epoll::wait to prevent UB in safe code

Twister915:
Fix to address issue #112

### DIFF
--- a/src/linux/epoll.rs
+++ b/src/linux/epoll.rs
@@ -265,8 +265,6 @@ impl Epoll {
     ///
     /// # Arguments
     ///
-    /// * `max_events` - the maximum number of events that we want to be returned in
-    ///                  `events` buffer.
     /// * `timeout` - specifies for how long the `epoll_wait` system call will block
     ///               (measured in milliseconds).
     /// * `events` - points to a memory area that will be used for storing the events
@@ -292,15 +290,10 @@ impl Epoll {
     ///         EpollEvent::new(EventSet::OUT, 4),
     ///     )
     ///     .unwrap();
-    /// let ev_count = epoll.wait(10, -1, &mut ready_events[..]).unwrap();
+    /// let ev_count = epoll.wait(-1, &mut ready_events[..]).unwrap();
     /// assert_eq!(ev_count, 1);
     /// ```
-    pub fn wait(
-        &self,
-        max_events: usize,
-        timeout: i32,
-        events: &mut [EpollEvent],
-    ) -> io::Result<usize> {
+    pub fn wait(&self, timeout: i32, events: &mut [EpollEvent]) -> io::Result<usize> {
         // Safe because we give a valid epoll file descriptor and an array of epoll_event structures
         // that will be modified by the kernel to indicate information about the subset of file
         // descriptors in the interest list. We also check the return value.
@@ -308,7 +301,7 @@ impl Epoll {
             epoll_wait(
                 self.epoll_fd,
                 events.as_mut_ptr() as *mut epoll_event,
-                max_events as i32,
+                events.len() as i32,
                 timeout,
             )
         })
@@ -364,7 +357,6 @@ mod tests {
     fn test_epoll() {
         const DEFAULT__TIMEOUT: i32 = 250;
         const EVENT_BUFFER_SIZE: usize = 128;
-        const MAX_EVENTS: usize = 10;
 
         let epoll = Epoll::new().unwrap();
         assert_eq!(epoll.epoll_fd, epoll.as_raw_fd());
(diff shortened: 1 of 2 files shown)


src/linux/epoll.rs
```diff
@@ -301,6 +301,7 @@ impl Epoll {
         timeout: i32,
         events: &mut [EpollEvent],
     ) -> io::Result<usize> {
```
jiangliu, line 303: How about returning an error instead of asserting?

src/linux/epoll.rs
```diff
@@ -301,6 +301,7 @@ impl Epoll {
         timeout: i32,
         events: &mut [EpollEvent],
```
acatangiu, line 302: Why do we keep `events` slice decoupled from `max_events`?

Is there a usability argument  against dropping `max_events` as an explicit parameter and just using `events.len()` in the code?

andreeaflorescu: The use case for which we keep `events.len()` decoupled from `max_events` is that `events` can be allocated once, and reused multiple times in different scenarios. Say we are initializing events with a capacity of 256, but in some cases we just want to return 10 events, and in others 256. With this is approach, `events` can be reused.

The other reason why we went on this path, is because we tried to keep `vmm-sys-util` as a minimal, safe wrapper of existing utilities, so we just keep the same parameters as the inner functions we are calling.

This is not really documented, so it can be a bit confusing. Sorry about that.

lauralt: There are indeed some unofficial links, e.g. https://copyconstruct.medium.com/the-method-to-epolls-madness-d9d2d6378642, that state the fact that `max_events` is the `events` len and in general this is how you will use `epoll_wait`, but I wouldn't add this restriction to our API. The error throwing fix seems enough to me.

acatangiu: The current API looks very un-rusty and very c-like (from what you said, that's probably by design).

I think the better UX is to use `slice::len()` since in Rust a slice is a view in a buffer consisting of idx and len. The `max_events` param is totally extraneous in a Rust function.

It all comes down to using the full power of Rust slices:
> Say we are initializing events with a capacity of 256, but in some cases we just want to return 10 events, and in others 256. With this is approach, events can be reused.

You could do that idiomatically:
```diff
- let ev_count = epoll.wait(10, -1, &mut ready_events[..]).unwrap();
+ let ev_count = epoll.wait(-1, &mut ready_events[..10]).unwrap();
```

acatangiu: Adding the extraneous param just increases chances of misuse like seen here.

Adding the explicit sanity check for `max_events` and throwing an error also obviously works but moves _mistakes_ discovery from compile-time to runtime.

These are just my 2 cents, don't read these comments as blockers to the PR.

andreeaflorescu: Sounds good :+1:

Twister915: I really like this suggestion to remove the argument alltogether. @lauralt if you want to allocate a large `events` block but only fill part of it, then you can just slice it as described by @acatangiu. Can I make this change to remove the `max_events` parameter?

lauralt: Yeah, I got it from the beginning, sorry if it wasn't that clear :D,  I was actually thinking if there could be use cases where you pass a sub-slice as described by @acatangiu, but `max_events` still could be smaller than the len of that subslice. Anyway, after more thinking, I agree the safest and the most rustacean solution is to remove that field. Yep, you can go ahead :+1:

src/linux/epoll.rs
Twister915: resolved by removing parameter

src/linux/epoll.rs
```diff
@@ -295,20 +295,15 @@ impl Epoll {
     /// let ev_count = epoll.wait(10, -1, &mut ready_events[..]).unwrap();
     /// assert_eq!(ev_count, 1);
     /// ```
-    pub fn wait(
-        &self,
-        max_events: usize,
-        timeout: i32,
-        events: &mut [EpollEvent],
-    ) -> io::Result<usize> {
+    pub fn wait(&self, timeout: i32, events: &mut [EpollEvent]) -> io::Result<usize> {
```
lauralt, line 298: `max_events` should be removed from the arguments list as well: https://github.com/rust-vmm/vmm-sys-util/blob/master/src/linux/epoll.rs#L268. Also the documentation example and the unit tests should be updated.

Twister915: Cool, doing that now. My local won't pass some unrelated unit tests, so sorry for pushing so many times haha. Just letting your CI take care of it. Unit tests may not work on M1 Mac Mini:

```
test rand::tests::test_timestamp_cycles ... FAILED
test rand::tests::test_xor_psuedo_rng_u32 ... FAILED
test unix::tempdir::tests::test_create_dir ... FAILED
test unix::tempdir::tests::test_create_dir_in ... FAILED
```

I'll open a different issue around this.

lauralt: No problem, would be nice if you can also add a CHANGELOG entry in the `Unreleased` section. Thanks!

Twister915: Done!

lauralt:
Looks good to me, can you also squash the commits?

jiangliu:
It's a break change, we need to bump the version number for next release.

andreeaflorescu:
> It's a break change, we need to bump the version number for next release.

@jiangliu We've been bumping the version when we prepare the release on crates.io. Are you suggesting to do that now?

Twister915:
Squashed

andreeaflorescu (approved):

jiangliu (approved):

PR #115, opened 2021-02-04
Don't assert that file size is unchanged for test_large_write_zeroes

connorkuehl:
When fallocate fails in File's implementation of write_zeroes_at,
the fallback behavior is to write zeroes with write_at(); however,
this requires extending the filesize. Remove the assertion that
expects the filesize to remain the same -- otherwise tests will
spuriously fail on a system where fallocate returns -EOPNOTSUPP.

Signed-off-by: Connor Kuehl <ckuehl@redhat.com>

### DIFF
--- a/src/linux/write_zeroes.rs
+++ b/src/linux/write_zeroes.rs
@@ -268,10 +268,6 @@ mod tests {
         for read in &readback[(0x1_8001 + 0x200)..SIZE] {
             assert_eq!(*read, NON_ZERO_VALUE);
         }
-
-        // It's ok to write zeroes after the end of the file, but the file size should not change.
-        assert!(f.write_all_zeroes_at(SIZE as u64 - 0x200, 0x400).is_ok());
-        assert_eq!(f.metadata().unwrap().len(), SIZE as u64);
     }
 
     #[test]


src/linux/write_zeroes.rs
```diff
@@ -123,6 +123,14 @@ impl WriteZeroesAt for File {
             return Ok(length);
         }
 
+        let requested_len = length;
+        let file_len = self.metadata().map(|m| m.len() as usize)?;
+        let write_limit = match file_len.checked_sub(offset as usize) {
+            None => return Ok(length),
```
lauralt, line 129: I'm not quite sure if it's okay to return `0k(length)` here instead of `Ok(0)`. From what I understand, if `fallocate()` is supported, even if `offset + length` > file size (and FALLOC_FL_KEEP_SIZE is set), `length` bytes are considered to be zeroed out. Here, on the fallback path, we are not actually zeroing anything. I'm probably missing something, so an explanation would be greatly appreciated.

I'm also wondering if, on the fallback path, the right approach is to emulate the fallocate size retention, instead of writing `length` zeroes and extending the file as well :-?.

connorkuehl: > I'm not quite sure if it's okay to return 0k(length) here instead of Ok(0). From what I understand, if fallocate() is supported, even if offset + length > file size (and FALLOC_FL_KEEP_SIZE is set), length bytes are considered to be zeroed out. Here, on the fallback path, we are not actually zeroing anything. I'm probably missing something, so an explanation would be greatly appreciated.

No, I think you're right; initially I was trying to make sure both branches behaved similarly with their return values, but I don't think there's a way to faithfully write those zeroes without extending the file size. I thought the file size was not supposed to change during this function call the way the unit test is written.

> I'm also wondering if, on the fallback path, the right approach is to emulate the fallocate size retention, instead of writing length zeroes and extending the file as well :-?.

I agree that seems more faithful to what fallocate does on-disk. The unit test suggests the file size is a property that shouldn't change as a result of this function's execution; is that not a real constraint here? I'm not very familiar with the intent behind that check.

lauralt: Nah, it was just me failing at writing unit tests. I was under the impression the file size won't change on the fallback path either. Thanks for reporting this! I think it should be enough to just remove that check. How does that sound?

connorkuehl: Thanks, @lauralt, I really appreciate it :-) I've pushed a new revision.

lauralt:
Looks good to me, @andreeaflorescu @jiangliu @sameo can you please take a look as well?

andreeaflorescu (approved):

jiangliu (approved):

Issue #104, opened 2020-10-15
Extend the FamStruct implementation to support multiple fields


andreeaflorescu:
The FAM struct implementation is minimalist and only works for FAM structs that have 2 fields: the number of entries and the incomplete array type.

FAM structs by their definition just require that the last field in the structure is the incomplete array type, but it does not talk about the number of other fields.

We should extend this implementation to support other scenarios as well. Because of this miss, vfio-ioctls cannot use the `FamStructWrapper` code and needs to redefine how to work with FAM.

Issue #99, opened 2020-10-08
FamStructWrapper: validate maximum capacity when doing new


andreeaflorescu:
The `new` function in `FamStructWrapper` is not validating that `num_elements` is less than the maximum capacity that the type supports (as defined in `FamStruct::max_len`), and thus might result in allocating more elements than it was initially intended.

A change in `FamStructWrapper` requires changes in `kvm-bindings` as well, so we might want to fix the vmm-sys-util dependency in kvm-bindings before doing a release. It might be better to remove the vmm-sys-util dependency from kvm-bindings (as this change does not pertain to the scope of kvm-bindings), and implement FamStruct directly in kvm-ioctls where it is actually needed.

Issue #85, opened 2020-04-15
FamStructWrapper: incorrect Clone implementation 


lauralt:
`Clone` implementation for `FamStructWrapper` keeps only the number of entries and the entries array members. This implementation should be either removed or modified so that all fields are cloned.

Twister915:
I'm trying to understand this issue, please correct me if I'm wrong. A FAM struct, in the typical case, just has length and a bunch of entries. Therefore, in this case, the `Clone` implementation is correct. However, FAM doesn't strictly need length + elements, but can have ... other fields ... + length + ... other fields ... + entries, and this is the case that `Clone` does not currently handle correctly. If that's correct, then I think if #104 is addressed first, then this should be pretty straight-forward.

andreeaflorescu:
@Twister915 I find it to be the other way around. We already support other fields in FAM Struct, but the clone is implemented such that only the `entries` and `length` are cloned. This makes the `Clone` implementation wrong, and as a consequence, the `FamStructWrapper` implementation also wrong (in the scenarios where clone is needed). Once we fix the clone, and add a few tests with FAM structs that have more fields, we can close both issues.

PR #118, opened 2021-02-17
fix fam issues

lauralt:
#85, #99, #104

### DIFF
--- a/src/fam.rs
+++ b/src/fam.rs
@@ -185,7 +185,15 @@ impl<T: Default + FamStruct> FamStructWrapper<T> {
     /// # Arguments
     ///
     /// * `num_elements` - The number of elements in the FamStructWrapper.
-    pub fn new(num_elements: usize) -> FamStructWrapper<T> {
+    ///
+    /// # Errors
+    ///
+    /// When `num_elements` is greater than the max possible len, it returns
+    /// `Error::SizeLimitExceeded`.
+    pub fn new(num_elements: usize) -> Result<FamStructWrapper<T>, Error> {
+        if num_elements > T::max_len() {
+            return Err(Error::SizeLimitExceeded);
+        }
         let required_mem_allocator_capacity =
             FamStructWrapper::<T>::mem_allocator_len(num_elements);
 
@@ -196,7 +204,7 @@ impl<T: Default + FamStruct> FamStructWrapper<T> {
         }
         mem_allocator[0].set_len(num_elements);
 
-        FamStructWrapper { mem_allocator }
+        Ok(FamStructWrapper { mem_allocator })
     }
 
     /// Create a new FamStructWrapper from a slice of elements.
@@ -205,15 +213,20 @@ impl<T: Default + FamStruct> FamStructWrapper<T> {
     ///
     /// * `entries` - The slice of [`FamStruct::Entry`](trait.FamStruct.html#associatedtype.Entry)
     ///               entries.
-    pub fn from_entries(entries: &[T::Entry]) -> FamStructWrapper<T> {
-        let mut adapter = FamStructWrapper::<T>::new(entries.len());
+    ///
+    /// # Errors
+    ///
+    /// When the size of `entries` is greater than the max possible len, it returns
+    /// `Error::SizeLimitExceeded`.
+    pub fn from_entries(entries: &[T::Entry]) -> Result<FamStructWrapper<T>, Error> {
+        let mut adapter = FamStructWrapper::<T>::new(entries.len())?;
 
         {
             let wrapper_entries = adapter.as_mut_fam_struct().as_mut_slice();
             wrapper_entries.copy_from_slice(entries);
         }
 
-        adapter
+        Ok(adapter)
     }
 
     /// Create a new FamStructWrapper from the raw content represented as `Vec<T>`.
@@ -417,7 +430,10 @@ impl<T: Default + FamStruct> PartialEq for FamStructWrapper<T> {
 
 impl<T: Default + FamStruct> Clone for FamStructWrapper<T> {
     fn clone(&self) -> Self {
-        FamStructWrapper::from_entries(self.as_slice())
+        // It is safe to unwrap() for now since `self` is a valid `FamStructWrapper`, so the number
+        // of entries can't be > T::max_len(). In the next commit, unwrap() will not be needed
+        // anymore.
+        FamStructWrapper::from_entries(self.as_slice()).unwrap()
     }
 }
 
@@ -470,6 +486,8 @@ where
             where
                 V: SeqAccess<'de>,
             {
+                use serde::de::Error;
+
                 let header = seq
                     .next_element()?
                     .ok_or_else(|| de::Error::invalid_length(0, &self))?;
@@ -477,7 +495,8 @@ where
                     .next_element()?
                     .ok_or_else(|| de::Error::invalid_length(1, &self))?;
 
-                let mut result: Self::Value = FamStructWrapper::from_entries(entries.as_slice());
+                let mut result: Self::Value = FamStructWrapper::from_entries(entries.as_slice())
+                    .map_err(|e| V::Error::custom(format!("{:?}", e)))?;
                 std::mem::replace(&mut result.mem_allocator[0], header);
                 Ok(result)
             }
@@ -638,7 +657,7 @@ mod tests {
     fn test_new() {
         let num_entries = 10;
 
-        let adapter = MockFamStructWrapper::new(num_entries);
+        let adapter = MockFamStructWrapper::new(num_entries).unwrap();
         assert_eq!(num_entries, adapter.capacity());
 
         let u32_slice = unsafe {
@@ -651,6 +670,15 @@ mod tests {
         for entry in u32_slice[1..].iter() {
             assert_eq!(*entry, 0);
         }
+
+        // It's okay to create a `FamStructWrapper` with the maximum allowed number of entries.
+        let adapter = MockFamStructWrapper::new(MockFamStruct::max_len()).unwrap();
+        assert_eq!(MockFamStruct::max_len(), adapter.capacity());
+
+        assert!(matches!(
+            MockFamStructWrapper::new(MockFamStruct::max_len() + 1),
+            Err(Error::SizeLimitExceeded)
+        ));
     }
 
     #[test]
@@ -662,7 +690,7 @@ mod tests {
             entries.push(i as u32);
         }
 
-        let adapter = MockFamStructWrapper::from_entries(entries.as_slice());
+        let adapter = MockFamStructWrapper::from_entries(entries.as_slice()).unwrap();
         let u32_slice = unsafe {
             std::slice::from_raw_parts(
                 adapter.as_fam_struct_ptr() as *const u32,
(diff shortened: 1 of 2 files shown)


lauralt:
@jiangliu was there any other concern regarding #104 besides the `Clone` and `PartialEq` implementations?
I've noticed in `vfio_ioctls` the following structure: 
```rust
struct vfio_region_info_with_cap {
    region_info: vfio_region_info,
    cap_info: __IncompleteArrayField<u8>,
}
```
which can be somehow seen as a `fam` struct even though the entries length is based on the `region_info` fields.

`generate_fam_struct_impl` indeed won't work for a structure that doesn't contain its own `len` field, but I don't think fixing this should be part of #104 since that macro is somehow of a helper (and I guess `FamStruct` can be implemented for the above structure as well). What do you think?

src/fam.rs
```diff
@@ -409,15 +422,50 @@ impl<T: Default + FamStruct> FamStructWrapper<T> {
...
     }
 }
 
 impl<T: Default + FamStruct> Clone for FamStructWrapper<T> {
     fn clone(&self) -> Self {
-        FamStructWrapper::from_entries(self.as_slice())
+        // The number of entries (self.as_slice().len()) can't be > T::max_len() since `self` is a
+        // valid `FamStructWrapper`.
+        let required_mem_allocator_capacity =
+            FamStructWrapper::<T>::mem_allocator_len(self.as_slice().len());
```
jiangliu, line 447: Should we clone the source FamStruct exactly by using "self.capacipy" instead of "self.as_slice().len()"?

src/fam.rs
```diff
@@ -422,9 +422,20 @@ impl<T: Default + FamStruct> FamStructWrapper<T> {
...
     fn eq(&self, other: &FamStructWrapper<T>) -> bool {
-        self.as_slice() == other.as_slice()
+        // This is safe as long as the requirements for the `FamStruct` trait to be safe are met
+        // (the implementing type and the entries elements are POD, therefore `Copy`, so memory
+        // safety can't be violated by the ownership of `fam_struct`s). It is also safe because
+        // we're trying to read `T`s from `&T`s that are pointing to properly initialized and
+        // aligned `T`s.
+        let (fam_struct, other_fam_struct): (T, T) = unsafe {
+            (
+                std::ptr::read(self.as_fam_struct_ref()),
```
jiangliu, line 434: why do we need to std::ptr::read() instead of comparing self.as_fam_struct_ref() and other.as_fam_struct_ref() directly?

src/fam.rs
```diff
@@ -938,4 +959,55 @@ mod tests {
...
+        generate_fam_struct_impl!(Foo, u32, entries, u16, length, 100);
+
+        type FooFamStructWrapper = FamStructWrapper<Foo>;
+
+        let mut foo = FooFamStructWrapper::new(0).unwrap();
+        foo.as_mut_fam_struct().index = 1;
+        foo.as_mut_fam_struct().flags = 2;
+        foo.as_mut_fam_struct().length = 3;
+        foo.push(3).unwrap();
+        foo.push(14).unwrap();
```
acatangiu, line 995: if `length` is 3 shouldn't we push 3 elements?

src/fam.rs
```diff
@@ -938,4 +959,55 @@ mod tests {
...
+        assert_eq!(foo.as_slice().len(), 3 + 2);
+        assert_eq!(foo.as_slice()[3 + 0], 3);
+        assert_eq!(foo.as_slice()[3 + 1], 14);
```
acatangiu, line 998: why is this 5? as slice should only provide the list of entries `&[T::entry]`, right? I was expecting len to be 2 since we started off with 0 and pushed only 2.

Documentation for `as_slice()` says it gets the elements/entries. In your tests it seems it also comes back with contents of FAM struct packed as elements.

I am now confused 😄

src/fam.rs
```diff
@@ -938,4 +959,55 @@ mod tests {
...
+        assert_eq!(
+            foo.as_mut_fam_struct().length,
+            foo2.as_mut_fam_struct().length
+        );
+        assert_eq!(
+            foo.as_mut_fam_struct().flags,
+            foo2.as_mut_fam_struct().flags
+        );
+        assert_eq!(foo.as_slice(), foo2.as_slice());
+        assert_eq!(foo2.as_slice().len(), foo2.as_mut_fam_struct().length as usize);
```
acatangiu: This also confuses me..

`foo.as_slice().len()` was checked to be `3 + 2`, then here `foo2.as_slice().len()` is checked to be `3` (`foo2.as_mut_fam_struct().length` is 3).

src/fam.rs
lauralt: Well, this was confusing for me as well, when you use `push()` it appends the element at the end of the entries collection which is considered as already having 3 elements (because of the above line). Anyway, this is not something that was added in this PR and I should've probably kept the tests simpler. Also, the whole `fam` module needs better documentation :D.

src/fam.rs
```diff
@@ -430,10 +430,31 @@ impl<T: Default + FamStruct> PartialEq for FamStructWrapper<T> {
 
 impl<T: Default + FamStruct> Clone for FamStructWrapper<T> {
     fn clone(&self) -> Self {
```
serban300, line 443: I think we can improve the previous implementation by also copying the header. Something like:

```
// It is safe to unwrap() for now since `self` is a valid `FamStructWrapper`, so the number
// of entries can't be > T::max_len().
let mut clone = FamStructWrapper::from_entries(self.as_slice()).unwrap();
clone.mem_allocator[0] = unsafe { std::ptr::read(self.as_fam_struct_ref()) };

clone
```

Would there be any concern related to this ?

lauralt: I want to avoid the `unwrap()` because if in the future `from_entries()` will return some other error as well, there's a high chance we're going to miss fixing this `unwrap()` and stuff will become risky. Also with your change, we are actually doing more things than we need.
I would also like to avoid the duplicate code that I added here, but didn't find a decent way to do that and not make things awkward.

src/fam.rs
```diff
@@ -938,4 +959,55 @@ mod tests {
         type Message2FamStructWrapper = FamStructWrapper<Message2>;
         assert!(serde_json::from_str::<Message2FamStructWrapper>(data_ser.as_str()).is_err());
     }
+
+    #[test]
+    fn test_clone_multiple_fields() {
```
serban300, line 976: It would be nice to also test the case where we drop the original structure and make sure that the clone remains working. Just to make sure that the new structure doesn't point to something in the old one. Normally it shouldn't happen, but we're doing some unsafe things.

src/fam.rs
```diff
@@ -430,10 +430,31 @@ impl<T: Default + FamStruct> PartialEq for FamStructWrapper<T> {
...
+            FamStructWrapper::<T>::mem_allocator_len(self.as_slice().len());
+
+        let mut mem_allocator = Vec::with_capacity(required_mem_allocator_capacity);
+
+        // This is safe as long as the requirements for the `FamStruct` trait to be safe are met
+        // (the implementing type and the entries elements are POD, therefore `Copy`, so memory
+        // safety can't be violated by the ownership of `fam_struct`). It is also safe because we're
+        // trying to read a T from a `&T` that is pointing to a properly initialized and aligned T.
+        unsafe {
+            let fam_struct: T = std::ptr::read(self.as_fam_struct_ref());
```
serban300, line 456: what if the FAM header contains a pointer ? Can it happen ?

src/fam.rs
lauralt: With your previous proposal, the things that were not needed would be:
- checking that the number of entries > max_len()
- pushing T::default() and at the end updating with the Fam value instead of directly pushing this value
- explicitly setting the `len` field of the Fam: `mem_allocator[0].set_len(num_elements);` which if I understand correctly, shouldn't be needed anymore if `len` is read from `self`

src/fam.rs
lauralt: Good question! As discussed offline, this might be possible, but `T` must be `Default` and from what I understand, implementing `Default` for a pointer becomes really awkward. Also, if a pointer is `Copy`, I wonder if `ptr::read` is actually unsafe :-?.
@jiangliu @alexandruag do you have some input on this? Thanks!

jiangliu: `FamStruct` is an unsafe trait, which requires following conditions for safety
```
/// Trait for accessing properties of C defined FAM structures.
///
/// This is unsafe due to the number of constraints that aren't checked:
/// * the implementer should be a POD
/// * the implementor should contain a flexible array member of elements of type `Entry`
/// * `Entry` should be a POD
```
So it's illegal for an FamStruct implementor to host pointer fields.

lauralt: I was under the same impression, but after some googling/asking around, looks like a pointer is a POD. So isn't this actually right? :-?

jiangliu: I like this definition from https://stackoverflow.com/questions/45634083/is-there-a-concept-of-pod-types-in-rust:
POD: Types whose values can be duplicated simply by copying bits.

serban300: Sorry, my bad. Now I remember that the entire class was written with the assumption that `T` is memcopyable. It's just very strange that in C++ the following code prints true:
```
struct Pod {
    int *a;
};

int main()
{
    cout<<is_pod<Pod>::value<<endl;

    return 0;
}
```

alexandruag: Hi! Don't know what authoritative source is best to quote here, but POD structs can contain pointers, which are effectively just integers meant to be interpreted as addresses. A pointer holding a value does not come with any guarantees by itself, and all safety concerns are passed on to whomever actually uses that pointer for a memory operation later (since all fields of a POD struct are always accessible and can be modified, any method operating on them must take into account that every possible value can be encountered). So, even if the POD struct contains pointers (which is unlikely), that doesn't seem to break any of the current assumptions.

aghecenco: I think the place to go would be the [C++ standard](https://github.com/cplusplus/draft). I don't own the hard copy, but the latest snapshot from the official repo loosely defines POD types through the template @serban300 used -

```
is_pod<T> is a Cpp17UnaryTypeTrait (20.15.2) with a base characteristic of true_type if T is a POD
type, and false_type otherwise. A POD class is a class that is both a trivial class and a standard-layout
class, and has no non-static data members of type non-POD class (or array thereof). A POD type is a
scalar type, a POD class, an array of such a type, or a cv-qualified version of one of these types.

[Note 1 : It is unspecified whether a closure type (7.5.5.2) is a POD type. — end note]
```

So my understanding here is in line with @alexandruag in the sense that scalar types are inherently POD, and since a pointer is basically just a number, it falls into this category.

I also found an unofficial definition that (more helpfully IMO) defines POD types as anything without a vtable.

serban300: Thanks for the clarifications. I agree, I don't have any concern here. I will resolve this comment.

src/fam.rs
lauralt: After having a chat with @alexandruag it turned out it would be better to avoid this kind of confusion and he suggested to add the `Clone` trait bound to `FamStruct`; this way we would simplify things a lot by simply cloning the `mem_allocator`. Does anyone have anything against this or see a potential problem with it?

jiangliu:
> @jiangliu was there any other concern regarding #104 besides the `Clone` and `PartialEq` implementations?
> I've noticed in `vfio_ioctls` the following structure:
> 
> ```rust
> struct vfio_region_info_with_cap {
>     region_info: vfio_region_info,
>     cap_info: __IncompleteArrayField<u8>,
> }
> ```
> 
> which can be somehow seen as a `fam` struct even though the entries length is based on the `region_info` fields.
> 
> `generate_fam_struct_impl` indeed won't work for a structure that doesn't contain its own `len` field, but I don't think fixing this should be part of #104 since that macro is somehow of a helper (and I guess `FamStruct` can be implemented for the above structure as well). What do you think?

When I was working on vfio-ioctls, I investigated the way to use generate_fam_struct_impl. But seems it's a little complex, so maybe it's ok to keep the hand-writing version for vfio-ioctls.

src/fam.rs
blitz: As a rule of thumb anything in C++ is POD that can be used from C. So any C struct is almost by definition POD.

Personally, I find [cppreference](https://en.cppreference.com/w/cpp/named_req/PODType) easier to read than the standard, but it's not authoritative.

jiangliu (approved):

alexandruag (approved):

PR #125, opened 2021-03-09
Unsafe blocks safety

iulianbarbu:
## Reason for this PR

There are ocurrences of unsafe blocks that do not have comments regarding why the operations are safe.

## Description of this PR

Added comments for why the unsafe blocks are safe and logic to back up the safety assurance.

### DIFF
--- a/src/linux/sock_ctrl_msg.rs
+++ b/src/linux/sock_ctrl_msg.rs
@@ -17,6 +17,8 @@ use crate::errno::{Error, Result};
 use libc::{
     c_long, c_void, cmsghdr, iovec, msghdr, recvmsg, sendmsg, MSG_NOSIGNAL, SCM_RIGHTS, SOL_SOCKET,
 };
+use std::os::raw::c_int;
+use crate::syscall::SyscallReturnCode;
 
 // Each of the following macros performs the same function as their C counterparts. They are each
 // macros because they are used to size statically allocated arrays.
@@ -203,14 +205,15 @@ fn raw_recvmsg(fd: RawFd, iovecs: &mut [iovec], in_fds: &mut [RawFd]) -> Result<
     let mut msg = new_msghdr(iovecs);
 
     if !in_fds.is_empty() {
+        // MSG control len is size_of(cmsghdr) + size_of(RawFd) * in_fds.len().
         msg.msg_control = cmsg_buffer.as_mut_ptr() as *mut c_void;
         set_msg_controllen(&mut msg, cmsg_capacity);
     }
 
     // Safe because the msghdr was properly constructed from valid (or null) pointers of the
     // indicated length and we check the return value.
+    // TODO: Should we handle MSG_TRUNC in a specific way?
     let total_read = unsafe { recvmsg(fd, &mut msg, 0) };
-
     if total_read == -1 {
         return Err(Error::last());
     }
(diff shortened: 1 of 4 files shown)


src/linux/timerfd.rs
```diff
@@ -200,6 +200,7 @@ impl AsRawFd for TimerFd {
 }
 
 impl FromRawFd for TimerFd {
+    // The safety for the function usage must be ensured by its users.
```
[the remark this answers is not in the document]
iulianbarbu: Done. PTAL!

src/linux/sock_ctrl_msg.rs
```diff
@@ -220,28 +220,42 @@ fn raw_recvmsg(fd: RawFd, iovecs: &mut [iovec], in_fds: &mut [RawFd]) -> Result<
...
+        // that it only happens when there is at least sizeof(cmsghdr) space after the pointer to
+        // read.
         let cmsg = unsafe { (cmsg_ptr as *mut cmsghdr).read_unaligned() };
 
         if cmsg.cmsg_level == SOL_SOCKET && cmsg.cmsg_type == SCM_RIGHTS {
             let fd_count = ((cmsg.cmsg_len - CMSG_LEN!(0)) as usize) / size_of::<RawFd>();
+            let mut end_idx = in_fds.len();
+
+            if copied_fds_count + fd_count >= end_idx {
+                is_full = true;
```
jiangliu, line 237: When is_full becomes true, should we close all un-copied RawFds? Otherwise those un-copied fds will get leaked.

iulianbarbu: I think we do not need to do this due to the fact that `cmsg_buffer` has the same capacity as `in_fds`. This means that when we receive ancillary data, we can hold at maximum `cmsg_buffer` capacity bytes, which corresponds as well to `in_fds` capacity, so copying from one to the other does not imply handling extra data. If the sender transmits more data than the receiver can hold, some data gets discarded. I've experimented with this in the added unit test. Please let me know if this makes sense.

iulianbarbu:
@jiangliu , please take a look now and let me know what you think.

src/linux/sock_ctrl_msg.rs
```diff
@@ -219,29 +220,42 @@ fn raw_recvmsg(fd: RawFd, iovecs: &mut [iovec], in_fds: &mut [RawFd]) -> Result<
         return Ok((0, 0));
     }
 
+    // Reference to a memory area with a CmsgBuffer, which contains a `cmsghdr` struct followed
```
jiangliu, line 223: Seems we should check MSG_TRUNC  and MSG_CTRUNC  in msghdr.msg_flags for safety.

iulianbarbu: Thanks for suggesting this. Indeed, checking for `MSG_CTRUNC` (`MSG_TRUNC` is not relevant for the scope of the PR) brought more light to how to assure the safety of this function. I've found out that there are two cases we need to be aware about (added unit tests for each):

- `raw_recvmsg` allocates enough data for receiving the ancillary data (within the `CmesgBuffer`) - it is computed based on the buffer received by the function from the outer scope, but it is also aligned, adding extra capacity which in these case accounts for few extra fds that can not be stored in the buffer from the outer scope. For this case I've added code paths to close the extra `fds`.
- `raw_recvmsg` does not allocate enough data for receiving the ancillary data and implied, the buffer from the outer scope as well, so the inner call to `recvmsg` sets the mshdr flag `MSG_CTRUNC` and drops any extra ancillary data. For this case we do not need to handle any extra data.

@jiangliu , please let me know if all of the above make sense to you.

src/linux/sock_ctrl_msg.rs
jiangliu: I'm hesitating about what to return in case of receiving unexpected data/fds. It's a sign of incorrect communication protocol, so it's a little risky to silently drop the unexpected data/fds. How about returning an error here?

src/linux/sock_ctrl_msg.rs
jiangliu: I'm hesitating about what to return in case of receiving unexpected data/fds. It's a sign of incorrect communication protocol, so it's a little risky to silently drop the unexpected data/fds. How about returning an error here?

iulianbarbu: Initially I've thought providing a best effort implementation, but as you pointed out, its a sign of incorrect communication protocol, and it would be safer to exclude it from the functionality altogether. With this picture in mind and for a clear expectation for the end users, I think it would be ideal to drop all the received fds from the control data - the ones that can be held in the buffer and the extra ones as well - before returning the error. WDYT?

jiangliu: exactly:)

jiangliu (approved):

andreeaflorescu (approved):

jiangliu (approved):

PR #135, opened 2021-08-12
sock_ctrl_msg: mark recv_with_fds as unsafe

alyssais:
Writes to arbitrary pointers are unsafe in Rust.  It's the caller's job to ensure that it's safe for the memory they are writing to can contain whatever arbitrary bytes are received over the socket.  For example, it would be unsafe to have an `iovec` pointing to the return value of `str::as_mut_ptr`, because strings can only contain byte sequences that are valid UTF-8.

Because it's on the caller to make sure they're passing pointers safely, any function that writes to `iovec`s has to be marked as unsafe.

Signed-off-by: Alyssa Ross <hi@alyssa.is>

***

Corresponding vhost change: https://github.com/rust-vmm/vhost/pull/69

### DIFF
--- a/src/linux/sock_ctrl_msg.rs
+++ b/src/linux/sock_ctrl_msg.rs
@@ -198,7 +198,11 @@ fn raw_sendmsg<D: IntoIovec>(fd: RawFd, out_data: &[D], out_fds: &[RawFd]) -> Re
     }
 }
 
-fn raw_recvmsg(fd: RawFd, iovecs: &mut [iovec], in_fds: &mut [RawFd]) -> Result<(usize, usize)> {
+unsafe fn raw_recvmsg(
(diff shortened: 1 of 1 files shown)


jiangliu (dismissed):

jiangliu (dismissed):

alyssais:
Do I need to add tests to get the coverage up like the CI says, even though line count has just increased because of reformatting?

jiangliu:
Only meaningful test cases are really needed:)
But really welcome for useful test cases!

alyssais:
I've rebased this in the hope that e39373fc6efc63b44f298633ba7f72f571224154 will fix the coverage issue.

jiangliu (approved):

lauralt (approved):

PR #136, opened 2021-08-13
tempdir: fix test assuming that TMPDIR=/tmp

alyssais:
I had TMPDIR set to /run/user/1000, and this made this test fail.

Signed-off-by: Alyssa Ross <hi@alyssa.is>

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 87.7,
+  "coverage_score": 87.6,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/unix/tempdir.rs
+++ b/src/unix/tempdir.rs
@@ -146,7 +146,7 @@ mod tests {
         let path = t.as_path();
         assert!(path.exists());
         assert!(path.is_dir());
-        assert!(path.starts_with("/tmp/"));
+        assert!(path.starts_with(temp_dir()));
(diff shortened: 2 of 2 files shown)


alyssais:
I don't really understand the coverage failure here, since all I've done is change a test?

andreeaflorescu:
It probably is a kcov bug, I can't figure it out why it fails either. Can you update the coverage nevertheless so that the tests passes? In the meantime we can look into why kcov thinks the coverage is decreased.

lauralt:
stderr says the following:
```


----------------------------- Captured stderr call -----------------------------
--
[src/linux/poll.rs:834] busy poll wait loop with hungup FDs detected on thread linux::poll::tests::test_poll_context_hungup

 
```
Can it be some intermittent failure or something like that?

alyssais:
> Can it be some intermittent failure or something like that?

No, that's intentional.  That test is marked `should_panic`, and the hangup detection code prints that message before panicking:

https://github.com/rust-vmm/vmm-sys-util/blob/0f740eb048579e951ae5565068059eec4d53a47c/src/linux/poll.rs#L826-L846

It would probably be better if it didn't print the other message if it was panicking, though.

alyssais:
@andreeaflorescu is that what you meant by updating the coverage?

lauralt (approved):

jiangliu (approved):

PR #159, opened 2022-03-18
Enhance rand to support macos

jiangliu:
### Summary of the PR

Enhance rand to support macos, also fixes some failure in unit tests on macos 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/unix/tempdir.rs
+++ b/src/unix/tempdir.rs
@@ -144,6 +144,7 @@ mod tests {
         let path = t.as_path();
         assert!(path.exists());
         assert!(path.is_dir());
+        #[cfg(not(target_os = "macos"))]
         assert!(path.starts_with(temp_dir()));
     }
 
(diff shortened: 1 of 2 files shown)


src/unix/tempdir.rs
```diff
@@ -144,6 +144,7 @@ mod tests {
         let path = t.as_path();
         assert!(path.exists());
         assert!(path.is_dir());
+        #[cfg(not(target_os = "macos"))]
```
andreeaflorescu, line 147: Can we instead check that the path is in the equivalent temp_dir for mac os? Otherwise this test is not very useful.

src/unix/tempdir.rs
```diff
@@ -185,12 +186,14 @@ mod tests {
...
```
andreeaflorescu, line 189: Can we add an equivalent check here instead?

andreeaflorescu (commented):
Nice to see support for macos being added to rust-vmm. Thanks Gerry 👍

mjt0k (commented):
FWIW, not all platforms have 64bit atomic types.  There's configuration to check: cfg(target_has_atomic = "64"). I just submitted an issue to omit compilation of src/metrics.rs if target does not have 64bit atomics. Examples of such platforms are armel, mipsel and (surprize!) x32 (tho I think this last one is due to lack of proper rust support there, since i386 does have 64bit atomics).

PR #164, opened 2022-07-11
`ioctl_expr` const fn

JonathanWoollett-Light:
### Summary of the PR

Using a const fn for `ioctl_expr` instead of a declarative macro.

When possible it is always better to use a const fn over a macro. They offer better code readability, better documentation and stronger typing offering better safety.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/linux/ioctl.rs
+++ b/src/linux/ioctl.rs
@@ -9,27 +9,31 @@
 //! Macros and functions for working with
 //! [`ioctl`](http://man7.org/linux/man-pages/man2/ioctl.2.html).
 
-use std::os::raw::{c_int, c_ulong, c_void};
+use std::os::raw::{c_int, c_uint, c_ulong, c_void};
 use std::os::unix::io::AsRawFd;
 
-/// Expression that calculates an ioctl number.
-///
+// The only reason
+// [_IOC](https://elixir.bootlin.com/linux/v5.10.129/source/arch/alpha/include/uapi/asm/ioctl.h#L40)
+// is a macro in C is because C doesn't have const functions, it is always better when possible to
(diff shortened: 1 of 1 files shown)


src/linux/ioctl.rs
```diff
@@ -9,27 +9,26 @@
...
-    ($dir:expr, $ty:expr, $nr:expr, $size:expr) => {
-        (($dir << $crate::ioctl::_IOC_DIRSHIFT)
-            | ($ty << $crate::ioctl::_IOC_TYPESHIFT)
-            | ($nr << $crate::ioctl::_IOC_NRSHIFT)
-            | ($size << $crate::ioctl::_IOC_SIZESHIFT)) as ::std::os::raw::c_ulong
-    };
+/// The only reason
+/// [_IOC](https://elixir.bootlin.com/linux/v5.10.129/source/arch/alpha/include/uapi/asm/ioctl.h#L40)
+/// is a macro in C is because C doesn't have const functions, it is always better when possible to
+/// use a const function over a macro in Rust.
```
andreeaflorescu, line 26: This should not be part of the public documentation IMO, it is an implementation detail. Can you update this to be a regular comment instead?

andreeaflorescu (commented):
LGTM, I'll approve once the documentation is updated. @JonathanWoollett-Light I am planning on doing a release for this crate. Should we try to get this in the next release as well?

src/linux/ioctl.rs
JonathanWoollett-Light: Changed it to a normal comment.

jiangliu (approved):

PR #168, opened 2022-08-19
Implementing `std::error::Error` for `fam::Error`

JonathanWoollett-Light:
### Summary of the PR

`vmm_sys_util::fam::Error` does not currently implement `std::error::Error`.

This introduces issues when attempting to write an error which contains this.

For example the below code is not possible without this trait implementation:
```rust
#[derive(Debug,thiserror::Error)]
pub enum MyError {
    #[error("My error A {0:?}")]
    A(#[from] vmm_sys_util::fam::Error),
    // ...
}
```

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/fam.rs
+++ b/src/fam.rs
@@ -20,7 +20,6 @@
 use serde::de::{self, Deserialize, Deserializer, SeqAccess, Visitor};
 #[cfg(feature = "with-serde")]
 use serde::{ser::SerializeTuple, Serialize, Serializer};
-#[cfg(feature = "with-serde")]
 use std::fmt;
 #[cfg(feature = "with-serde")]
 use std::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


src/fam.rs
```diff
@@ -32,6 +31,14 @@ pub enum Error {
     /// The max size has been exceeded
     SizeLimitExceeded,
 }
+impl std::error::Error for Error {}
+impl fmt::Display for Error {
```
lauralt, line 35: nit: can you add newlines between these lines? Helps a bit with readability.
nit2: It would also be nice to mention in the commit message that you are implementing Display because of the Error trait bounds.

lauralt (approved):

src/fam.rs
JonathanWoollett-Light: Added newlines and updated commit message. Fixed?

lauralt (approved):

jiangliu (approved):

PR #174, opened 2022-10-04
Added function to generate 'N' random bytes

vibharya:
Signed-off-by: Vibha Acharya <vibharya@amazon.co.uk>

### Summary of the PR

Under rand there is only one function to generate 'n' alphanumeric characters.
Generating 'n' random bytes is a useful and generic functionality.

Also corrected typo in existing function names.  

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -66,6 +66,10 @@ fn xor_psuedo_rng_u8_alphanumerics(rand_fn: &dyn Fn() -> u32) -> Vec<u8> {
     r
 }
 
+fn xor_pseudo_rng_u8_bytes(rand_fn: &dyn Fn() -> u32) -> Vec<u8> {
+    rand_fn().to_ne_bytes().to_vec()
+}
+
 fn rand_alphanumerics_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> OsString {
     let mut buf = OsString::new();
     let mut done = 0;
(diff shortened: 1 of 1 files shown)


src/rand.rs
```diff
@@ -39,7 +39,7 @@ pub fn timestamp_cycles() -> u64 {
 }
 
 /// Generate pseudo random u32 numbers based on the current timestamp.
-pub fn xor_psuedo_rng_u32() -> u32 {
+pub fn xor_pseudo_rng_u32() -> u32 {
```
andreeaflorescu, line 42: Can you add this in a separate commit and specify that it fixes a typo in the function name? This will be a breaking change that we need to specify in the CHANGELOG, and otherwise it's easy to miss them.

src/rand.rs
```diff
@@ -80,10 +89,29 @@ fn rand_alphanumerics_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> OsString {
     }
 }
 
+fn rand_bytes_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> Vec<u8> {
+    let mut buf: Vec<u8> = Vec::new();
```
andreeaflorescu, line 93: The code in this module looks a bit C-like, and should probably be refactored in a future iteration, but there is no reason to continue the trend for new functions as well. Can you try to use more idiomatic ways of initializing the vector? For example, you can combine `iter` with `take` to retrieve the number of elements you need from a vector. Alternatively, you check how many times you have to call the `xor_pseudo_rng..` function and use a combination of `iter`, `map`, and `zip`.

src/rand.rs
vibharya: Sure, I'll add it in a separate commit in the same PR.

src/rand.rs
```diff
@@ -66,11 +66,20 @@ fn xor_psuedo_rng_u8_alphanumerics(rand_fn: &dyn Fn() -> u32) -> Vec<u8> {
     r
 }
 
+fn xor_pseudo_rng_u8_bytes(rand_fn: &dyn Fn() -> u32) -> Vec<u8> {
+    let mut r = vec![];
+
+    for n in &rand_fn().to_ne_bytes() {
+        r.push(*n);
+    }
+    r
```
andreeaflorescu, line 75: This can also be written as an 1-liner: `rand_fn().to_ne_bytes().to_vec()`

src/rand.rs
andreeaflorescu: Thanks! Can you please give more details in your commit message? Specify which which functions you renamed to what.

andreeaflorescu: The public function is mandatory because we need to include it in the changelog. You can also specify the other ones.

src/rand.rs
vibharya: Aah yes!

aghecenco (approved):

Issue #178, opened 2022-11-15
`rand_bytes()` causes a panic when `len` is not a multiple of 4 


vibharya:
Error/Panic message: `thread 'main' panicked at 'attempt to subtract with overflow'`

PR #179, opened 2022-11-15
Fix panic caused by rand_bytes()

vibharya:
Signed-off-by: Vibha Acharya <vibharya@amazon.co.uk>

### Summary of the PR

rand_bytes() causes a panic when number of bytes requested are not multiples of 4.
`'main' panicked at 'attempt to subtract with overflow'`

Relates to : #178

The unit test was testing for 8 which is why it passed.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## Unreleased
+
+### Changed
+- [[#178](https://github.com/rust-vmm/vmm-sys-util/issues/178)]: Fixed a bug in
+  `rand_bytes` that was triggering a panic when the number of bytes was not a
+  multiple of 4.
 
 ## v0.11.0
 
--- a/src/rand.rs
+++ b/src/rand.rs
@@ -81,10 +81,10 @@ fn rand_alphanumerics_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> OsString {
 
 fn rand_bytes_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> Vec<u8> {
     let mut buf: Vec<Vec<u8>> = Vec::new();
-    let mut num = len;
+    let mut num = if len % 4 == 0 { len / 4 } else { len / 4 + 1 };
     while num > 0 {
         buf.push(xor_pseudo_rng_u8_bytes(rand_fn));
-        num -= 4;
+        num -= 1;
     }
     buf.into_iter().flatten().take(len).collect()
 }
@@ -156,7 +156,8 @@ mod tests {
 
     #[test]
     fn test_rand_bytes() {
-        let s = rand_bytes(8);
-        assert_eq!(8, s.len());
+        for i in 0..8 {
+            assert_eq!(i, rand_bytes(i).len());
+        }
     }
 }


src/rand.rs
```diff
@@ -156,7 +156,7 @@ mod tests {
 
     #[test]
     fn test_rand_bytes() {
-        let s = rand_bytes(8);
-        assert_eq!(8, s.len());
+        let s = rand_bytes(7);
```
andreeaflorescu, line 159: Let's add multiple test scenarios here. Can you instead do a for from 0 to 8 and check that rand_bytes(i).len() is i?

andreeaflorescu (changes requested):
Please update the changelog to specify the bug you fixed. This is now part of the requirements for a PR.

andreeaflorescu:
Can you also add the following to your commit message:
```Fixes: https://github.com/rust-vmm/vmm-sys-util/issues/178```

File: src/rand.rs (hunk 1 above)
Comment (vibharya) [reply src/rand.rs:159]: Absolutely!

**Comment by andreeaflorescu** [comment]
> Please update the changelog to specify the bug you fixed. This is now part of the requirements for a PR.

There is still the changelog update needed.

File: CHANGELOG.md [hunk 2]
```diff
@@ -1,4 +1,11 @@
 # Changelog
+## Unreleased
+
+### Added
+
+### Changed
+- [[#178](https://github.com/rust-vmm/vmm-sys-util/issues/178)]: Fixed the
```
andreeaflorescu, line 7: Let's be more explicit about what was happening: fixed a bug in `rand_bytes` that was triggering a panic when the number of bytes was not a multiple of 4.

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 # Changelog
+## Unreleased
+
+### Added
```
andreeaflorescu, line 4: No need to add this as an empty section. When we're doing the release ideally the only change needed should be to replace Unreleased -> v0.x.y

andreeaflorescu (approved):

russell-islam (approved):

src/rand.rs
```diff
@@ -81,10 +81,10 @@ fn rand_alphanumerics_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> OsString {
 
 fn rand_bytes_impl(rand_fn: &dyn Fn() -> u32, len: usize) -> Vec<u8> {
     let mut buf: Vec<Vec<u8>> = Vec::new();
-    let mut num = len;
+    let mut num = if len % 4 == 0 { len / 4 } else { len / 4 + 1 };
```
jiangliu, line 84: How about 
```
let mut num. = (len + 3) >> 2;
```

andreeaflorescu: I find the if to be a bit easier to read 😬

aghecenco (approved):

src/rand.rs
arunpkm: Performance is more important than readability. So, `let mut num. = (len + 3) >> 2;` should have been opted. 

Refer Perplexity AI thread: https://www.perplexity.ai/search/what-is-the-compute-time-diffe-VJ_hoMTVT92_RlvS5Jd9sg 

```
Key Observations
Option 2 is consistently faster: For all tested inputs, the bit-shifting approach (len + 3) >> 2 outperforms the conditional approach if len % 4 == 0 { len / 4 } else { len / 4 + 1 }.

Performance improvement: Option 2 is approximately 23-35% faster than Option 1, with an average improvement of about 27%.

Consistency across inputs: The performance advantage holds regardless of whether the input is divisible by 4 or not.
```

PR #188, opened 2023-04-16
Disable compilation of Atomic64 code in metric.rs if there's no Atomic64

mjt0k:
Atomic64 is not present, at least, on armel, mipsel and x32 platforms.  Disable compilation of metrics.rs which enhances this data type.

### DIFF
--- a/src/metric.rs
+++ b/src/metric.rs
@@ -11,6 +11,8 @@
 //! of these components can choose what metrics they’re interested in and also
 //! can add their own custom metrics without the need to maintain forks.
 
+#![cfg(target_has_atomic = "64")]
+
 use std::sync::atomic::{AtomicU64, Ordering};
 
 /// Abstraction over the common metric operations.


src/metric.rs
```diff
@@ -11,6 +11,8 @@
 //! of these components can choose what metrics they’re interested in and also
 //! can add their own custom metrics without the need to maintain forks.
 
+#![cfg(target_has_atomic = "64")]
```
JonathanWoollett-Light, line 14: This would result in a compilation error when this condition is not true.

To implement this, it should be on https://github.com/rust-vmm/vmm-sys-util/blob/main/src/lib.rs#L21 like:

```rust
#[cfg(target_has_atomic = "64")]
pub mod metric;
```

mjt0k: Well, it doesn't result in compilation error: see https://buildd.debian.org/status/package.php?p=rust-vmm-sys-util&suite=sid - it compiled fine on all platforms where previously this caused an error.  I guess it will result in compilation error of crates which *use* this one, as there will be no implementation for the declared functionality.

But I must admit: I know *nothing* about rust. It's the 2nd day since I looked at what rust actually *is*, what I suggest is just something I've seen being used elsewhere. I've no idea if this is a correct solution or not.

andreeaflorescu:
@mjt0k can we think about a way to test this in the CI? I find it nice that we're adding support for other platforms as well, but the problem I am seeing is that (similar to the big endian support in vm-virtio) we are going to fix it, and future PRs will break it because we're not testing it.

src/metric.rs
andreeaflorescu: Writing `#![cfg(target_has_atomic = "64")]` (notice the `!`) is equivalent to:
```rust
#[cfg(target_has_atomic = "64")] pub mod metric;
```

We can still have it as @JonathanWoollett-Light suggested even if they're equivalent because it is making it a bit more obvious that the whole module is excluded from compilation on targets that don't support atomic64.

andreeaflorescu:
@mjt0k can you fix your commit message so we can merge this? We are expecting the summary to be less than 60 chars, and the body to be less than 72 (i think). We have some issues with the CI (cargo audit is not installed on arm, and that's why it's failing), so we can ignore that for now.

PR #189, opened 2023-04-16
linux/timerfd: stop explicitly casting to c_long for timespec.tv_nsec

mjt0k:
struct timespec.tv_nsec is implementation-specific (even if the doc says it is c_long). For example, on x32 it is i64.  Do not explicitly cast nsec value to c_long, use whatever type needed for the target type.  This also simplifies the code (avoiding temporary variable).

### DIFF
--- a/src/linux/timerfd.rs
+++ b/src/linux/timerfd.rs
@@ -71,9 +71,7 @@ impl TimerFd {
         {
             spec.it_value.tv_sec = dur.as_secs() as libc::time_t;
         }
-        // nsec always fits in i32 because subsec_nanos is defined to be less than one billion.
-        let nsec = dur.subsec_nanos() as i32;
-        spec.it_value.tv_nsec = libc::c_long::from(nsec);
+        spec.it_value.tv_nsec = dur.subsec_nanos() as _;
 
         if let Some(int) = interval {
             // https://github.com/rust-lang/libc/issues/1848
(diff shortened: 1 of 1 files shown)


src/linux/timerfd.rs
```diff
@@ -71,19 +71,15 @@ impl TimerFd {
         {
             spec.it_value.tv_sec = dur.as_secs() as libc::time_t;
         }
-        // nsec always fits in i32 because subsec_nanos is defined to be less than one billion.
-        let nsec = dur.subsec_nanos() as i32;
-        spec.it_value.tv_nsec = libc::c_long::from(nsec);
+        spec.it_value.tv_nsec = dur.subsec_nanos() as _;
```
JonathanWoollett-Light, line 74: Personal preference but I find `as _` to be quite unreadable.

```suggestion
        spec.it_value.tv_nsec = dur.subsec_nanos().try_into().unwrap();
```

src/linux/timerfd.rs
```diff
@@ -71,19 +71,15 @@ impl TimerFd {
...
         if let Some(int) = interval {
             // https://github.com/rust-lang/libc/issues/1848
             #[cfg_attr(target_env = "musl", allow(deprecated))]
             {
                 spec.it_interval.tv_sec = int.as_secs() as libc::time_t;
             }
-            // nsec always fits in i32 because subsec_nanos is defined to be less than one billion.
-            let nsec = int.subsec_nanos() as i32;
-            spec.it_interval.tv_nsec = libc::c_long::from(nsec);
+            spec.it_interval.tv_nsec = int.subsec_nanos() as _;
```
JonathanWoollett-Light, line 82: Personal preference but I find as _ to be quite unreadable.
```suggestion
        spec.it_value.tv_nsec = dur.subsec_nanos().try_into().unwrap();
```

mjt0k: Once again, this is 2nd day since I looked what rust is, to begin with, so I know nothing about the right constructs required here. I'm fine as long as it works. "My" version has been suggested by someone on IRC, - I didn't even know rust requires explicit type conversion between two different integer types.

FWIW, this issue happens on x32 *only*. On all the other platforms, timespec.tv_nsec is of long datatype. It is *documented* as being long. But on x32 it is int64_t. Apparently this is a bug in the architecture specifications.  Also, x32 is a very rare thing.

Still, I like the code without explicit integer conversion more. It might also be a good idea to retain the comment which was previously there.

PR #201, opened 2023-08-23
Update SyscallReturnCode to support types other than i32.

JBYoshi:
### Summary of the PR

Currently, `SyscallReturnCode` is limited to system call return values of type `i32`. However, not all system calls return 32-bit integer values. For example, the [`read` system call](https://man7.org/linux/man-pages/man2/read.2.html) returns an `ssize_t` type, which will map to a 64-bit integer on 64-bit systems.

This PR updates `SyscallReturnCode` to be a generic type that can contain any signed integer value, so it can be used with other integer types. Since existing code may not specify generic type values, I set the type parameter to default to the existing type of `c_int` to try and preserve compatibility.

### Requirements

requirements:

  description line.
  test.
- [ ] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/src/syscall.rs
+++ b/src/syscall.rs
@@ -7,12 +7,12 @@ use std::os::raw::c_int;
 
 /// Wrapper to interpret syscall exit codes and provide a rustacean `io::Result`.
 #[derive(Debug)]
-pub struct SyscallReturnCode(pub c_int);
+pub struct SyscallReturnCode<T: From<i8> + Eq = c_int>(pub T);
 
-impl SyscallReturnCode {
+impl<T: From<i8> + Eq> SyscallReturnCode<T> {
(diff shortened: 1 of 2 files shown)


src/syscall.rs
```diff
@@ -7,12 +7,12 @@ use std::os::raw::c_int;
 
 /// Wrapper to interpret syscall exit codes and provide a rustacean `io::Result`.
 #[derive(Debug)]
-pub struct SyscallReturnCode(pub c_int);
+pub struct SyscallReturnCode<T: From<i8> + Eq = c_int>(pub T);
```
roypat, line 10: Nice!

roypat (approved):

JonathanWoollett-Light (approved):

PR #208, opened 2023-11-01
Fix possible memory leak

JonathanWoollett-Light:
### Summary of the PR

If an error is returned after calling `CString::into_raw` and before calling `CString::from_raw` the memory allocated for the string is leaked. This could previously occur within `TempFile::new_with_prefix`.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/tempfile.rs
+++ b/src/tempfile.rs
@@ -61,22 +61,21 @@ impl TempFile {
         let mut os_fname = prefix.as_ref().to_os_string();
         os_fname.push("XXXXXX");
 
-        let raw_fname = match CString::new(os_fname.as_bytes()) {
-            Ok(c_string) => c_string.into_raw(),
-            Err(_) => return Err(Error::new(libc::EINVAL)),
+        let c_tempname = CString::new(os_fname.as_bytes()).map_err(|_| Error::new(libc::EINVAL))?;
+        let raw_tempname = c_tempname.into_raw();
+
+        // SAFETY: Safe because `c_tempname` is a null-terminated string, as it originates from
+        // `CString::into_raw`.
+        let ret = unsafe { libc::mkstemp(raw_tempname) };
+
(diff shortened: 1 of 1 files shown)


src/tempfile.rs
```diff
@@ -61,22 +61,20 @@ impl TempFile {
...
+        let c_tempname = CString::new(os_fname.as_bytes()).map_err(|_| Error::new(libc::EINVAL))?;
+        let raw_tempname = c_tempname.into_raw();
 
-        // SAFETY: Safe because `raw_fname` originates from CString::into_raw, meaning
-        // it is a pointer to a nul-terminated sequence of characters.
-        let fd = unsafe { libc::mkstemp(raw_fname) };
-        if fd == -1 {
-            return errno_result();
-        }
+        // SAFETY: Safe because `c_tempname` is a null-terminated string.
```
roypat, line 67: Why shorten this comment?

JonathanWoollett-Light: The safety of `mkstemp` only requires it is a null-terminated string. The original comment mentioning it was from `CString` was unnecessary. Removing this focuses more on the real reason.

roypat: Well, "`c_tempname` is a null-terminated string" is simply rephrasing the invariant that needs to be upheld, but does not tell us _why_ it is upheld. But that's arguable the more important part, and that's why we had the second part of the comment. Can you readd it, please?

JonathanWoollett-Light: I think safety comments only need to describe the immediate condition that being true makes it safe.

Otherwise the comments would need to describe the full trace of the program and how it arrived at this point with these values, or you make a subjective call about how much history you want to include.

roypat (approved):

JonathanWoollett-Light:
Fixes https://github.com/rust-vmm/vmm-sys-util/issues/212

PR #215, opened 2024-01-02
Make `as_mut_fam_struct()` public and unsafe

bchalios:
### Summary of the PR

We previously made as_mut_fam_struct() of FamStructWrapper private so
that we avoid that users modify the length of the header under our feet.
However, this disallows callers to modify other parts of the header
(which is totally safe to do).

Make as_mut_fam_struct() public again, but also mark it as unsafe,
asking users to guarantee that they will not change the length stored
within the header directly, via the mutable reference.

Also, prepare for release v0.12.1

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.12.1
+
+### Changed
+- [[#215](https://github.com/rust-vmm/vmm-sys-util/pull/215)]: Make
+  `as_mut_fam_struct()` public and unsafe to let users modify fields of the
+  header other than the length.
+
 ## v0.12.0
 
 ### Changed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "vmm-sys-util"
-version = "0.12.0"
+version = "0.12.1"
 authors = ["Intel Virtualization Team <vmm-maintainers@intel.com>"]
 description = "A system utility set"
 repository = "https://github.com/rust-vmm/vmm-sys-util"
--- a/src/fam.rs
+++ b/src/fam.rs
@@ -50,6 +50,8 @@ impl fmt::Display for Error {
 /// * the implementer should be a POD
 /// * the implementor should contain a flexible array member of elements of type `Entry`
 /// * `Entry` should be a POD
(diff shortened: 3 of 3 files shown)


roypat (approved):

JonathanWoollett-Light (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.12.1
+
+### Changed
+- [[#215](https://github.com/rust-vmm/vmm-sys-util/pull/215)]: Make
+  `as_mut_fam_struct()` public and unsafe to let users modify fields of the
+  header other than the length.
```
Ablu, line 8: Shouldn't this have been a v0.13.0 since an access was marked unsafe?

roypat: The function was made private in 0.12.0, so its essentially just adding a new function.

stefano-garzarella: IIUC in v0.12.0 `as_mut_fam_struct()` was marked private by mistake. Some users (e.g. vhost crate) used that API from previous versions, so IMHO v0.12.1 should be fine.

Maybe we should yank v0.12.0 since we did not want to make `as_mut_fam_struct()` not accessible.

Ablu: Ah. Sorry, missed that the removal was done in 0.12.0. Makes sense! I do not think a yank would have been required in that case.

stefano-garzarella: > I do not think a yank would have been required in that case.

@Ablu why not?
For example, `vhost` crate broke after updating from v0.11.0 to v0.12.0 because it used `as_mut_fam_struct()`.

Ablu: > For example, `vhost` crate broke after updating from v0.11.0 to v0.12.0 because it used `as_mut_fam_struct()`.

Yes, but that is "fine" as it was a major release which is allowed to break stuff. So if someone would have updated to `0.12.0` (so someone who does not use `as_mut_fam_struct()`), they would have been fine and would not necessarily need to be forced to `0.12.1`. Of course it is unlikely that this happened :)

stefano-garzarella: Okay, now I see your point ;-)

PR #221, opened 2024-04-17
fix: use random temporary directory in test

mgeisler:
### Summary of the PR

While trying to enable the unit tests in Android (see https://r.android.com/3044134), I noticed test failures saying:

    thread 'errno::tests::test_errno' panicked at
    external/rust/crates/vmm-sys-util/src/errno.rs:156:14:
    called `Result::unwrap()` on an `Err` value: Os { code: 21, kind:
    IsADirectory, message: "Is a directory" }

I believe this is because of the hard-coded name for the temporary file in the `test_errno` unit test.

Using a randomized temp directory should fix this.

### Requirements

requirements:

  test.

### DIFF
--- a/src/errno.rs
+++ b/src/errno.rs
@@ -132,7 +132,7 @@ pub fn errno_result<T>() -> Result<T> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::env::temp_dir;
+    use crate::tempfile::TempFile;
     use std::error::Error as _;
     use std::fs::OpenOptions;
     use std::io::{self, Read};
@@ -145,14 +145,16 @@ mod tests {
         let expected_errno = libc::EIO;
 
         // try to read from a file without read permissions
-        let mut path = temp_dir();
-        path.push("test");
+        let temp_file = TempFile::new().unwrap();
+        let path = temp_file.as_path().to_owned();
+        // Drop temp_file so we can cleanly reuse path below.
+        std::mem::drop(temp_file);
         let mut file = OpenOptions::new()
             .read(false)
             .write(true)
             .create(true)
             .truncate(true)
-            .open(path)
+            .open(&path)
             .unwrap();
         let mut buf: Vec<u8> = Vec::new();
         assert!(file.read_to_end(&mut buf).is_err());
@@ -174,6 +176,8 @@ mod tests {
         let last_err: io::Error = last_err.into();
         // Test creating a `std::io::Error` from an `Error`
         assert_eq!(io::Error::last_os_error().kind(), last_err.kind());
+
+        assert!(std::fs::remove_file(&path).is_ok());
     }
 
     #[test]


mgeisler:
> Using a randomized temp directory should fix this.

Right after implementing this, I noticed https://github.com/rust-vmm/vmm-sys-util/blob/main/src/tempfile.rs :slightly_smiling_face:  I don't really know the crate here, so would it be better to use the functionality there?

src/errno.rs
```diff
@@ -145,8 +144,8 @@ mod tests {
         let expected_errno = libc::EIO;
 
         // try to read from a file without read permissions
-        let mut path = temp_dir();
-        path.push("test");
+        let temp_dir = tempfile::tempdir().unwrap();
+        let path = temp_dir.path().join("test");
```
roypat, line 148:
```suggestion
        let temp_file = TempFile::new().unwrap();
        let path = temp_file.as_path(); // Can probably inline this into `.open()` below, but GitHub doesnt let me extend the comment that far
```

roypat (commented):
Hi!
Thanks for this PR! I agree that the failure you're seeing it because of the hardcoded path in the test. Probably something else on the system created a `test` directory inside `/tmp`, and so the test fails when it tries to create `test` as a file. (assuming that no one created something called "test" inside of `/tmp` is indeed bold of this test, haha) 

As for `tempfile`, let use the module from vmm-sys-util. We try to keep the dependencies on these crates minimal, so since its an easy change to use our module here, that'd be preferred.

mgeisler:
> Hi! Thanks for this PR! I agree that the failure you're seeing it because of the hardcoded path in the test. Probably something else on the system created a `test` directory inside `/tmp`, and so the test fails when it tries to create `test` as a file. (assuming that no one created something called "test" inside of `/tmp` is indeed bold of this test, haha)

Yeah, it's the kind of hidden assumption that works fine for a long time, until it doesn't because someone has an unexpected setup :smile: 

> As for `tempfile`, let use the module from vmm-sys-util. We try to keep the dependencies on these crates minimal, so since its an easy change to use our module here, that'd be preferred.

Alright! I applies your suggestion to use `TempFile` from this crate, but tweaked them a bit. I think using the temporary file directly does something slightly different, but I'm not sure if it matters? Before the `test` file did not exist when `open()` was called. If we use the `temp_file` directly, the file exists — and I guess it's a bit of a coincidence that the flags match up?

To avoid reasoning about this, I now simply delete the `temp_file` early and then reuse the `path` (with the assumption that nobody is trying to race me).

Please let me know if you think this is overkill?

roypat:
> > As for `tempfile`, let use the module from vmm-sys-util. We try to keep the dependencies on these crates minimal, so since its an easy change to use our module here, that'd be preferred.
> 
> Alright! I applies your suggestion to use `TempFile` from this crate, but tweaked them a bit. I think using the temporary file directly does something slightly different, but I'm not sure if it matters? Before the `test` file did not exist when `open()` was called. If we use the `temp_file` directly, the file exists — and I guess it's a bit of a coincidence that the flags match up?

I also don't think it really matter, as long as we do _something_ that causes a well-known errno to be set (e.g. reading from a file opened with `O_WRONLY` resulting in EBADF). 

> To avoid reasoning about this, I now simply delete the `temp_file` early and then reuse the `path` (with the assumption that nobody is trying to race me).
> 
> Please let me know if you think this is overkill?

No, this is perfectly fine, thanks for fixing this!

roypat (approved):

mgeisler:
Thanks, I'm glad you like it! The tests will then soon also be running as part of the Android build: https://r.android.com/3044134.

JonathanWoollett-Light (approved):

PR #233, opened 2025-01-09
Use portable-atomic when AtomicU64 is not available.

### LINKED ISSUE #187: Allow building on 32bit platforms without AtomicU64
Hi!

This code has src/metric.rs, which unconditionally uses std::sync::atomic::AtomicU64. Quite some platforms don't have 64-bit atomics with current rust, namely it is armel, mipsel, powerpc and x32 (heh).  Maybe the metric module can be guarded with the right #[cfg target_has_atomic] ?

Trying to package some stuff for Debian, this crate come as a dependency...

Thanks!

flaviojs:
### Summary of the PR

Use portable-atomic when AtomicU64 is not available.
Fixes #187

I do not know if AtomicU64 is required, or why other atomic types don't implement Metric.
If it is not required, an alternate change is to use the more portable AtomicUsize.

It is unclear if the changes to test `linux::sock_ctrl_msg::tests::buffer_len`are appropriate since I don't use this crate.

All tests pass in my 32-bit OS except the doc-tests `linux::ioctl::ioctl` and `linux::ioctl::ioctl_with_val`.
These failures are expected because I don't have `/dev/kvm`.

I do not know what should go into the changelog so I left it unchanged.

### Requirements

requirements:

  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ libc = "0.2.39"
 serde = { version = "1.0.27", optional = true }
 serde_derive = { version = "1.0.27", optional = true }
 
+[target.'cfg(not(target_has_atomic = "u64"))'.dependencies]
+portable-atomic = "1"
+
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
 bitflags = "1.0"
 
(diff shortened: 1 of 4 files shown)


rbradford:
Thank you for for your contribution but I don't think we support any 32-bit architectures in the rust-vmm ecosystem.

PR #242, opened 2025-06-26
Add EventNotifier as an abstraction for EventFd and PipeFd

uran0sH:
EventFd is linux specific. We need to fallback eventfd to pipefd on macOS and *BSD. Add an abstraction layer to support both eventfd and pipefd.

### Summary of the PR

*Please summarize here why the changes in this PR are needed.*

### Requirements

requirements:

- [ ] All commits in this PR have Signed-Off-By trailers (with
- [ ] All added/changed functionality has a corresponding unit/integration
  test.
- [ ] Any newly added `unsafe` code is properly documented.


src/event_notifier.rs
```diff
@@ -0,0 +1,291 @@
...
+            // our process and we handle the error case.
+            let ret = unsafe { libc::eventfd(0, flag) };
+            if ret < 0 {
+                Err(io::Error::last_os_error())
+            } else {
+                Ok(EventNotifier {
+                    // SAFETY: This is safe because we checked ret for success and know
+                    // the kernel gave us an fd that we own.
+                    rfd: unsafe { File::from_raw_fd(ret) },
+                    wfd: unsafe { File::from_raw_fd(ret) },
```
roypat, line 63: This won't work, as the `Drop` impls for both `File`s will try to close the FD `ret`, but the `File` drop impl aborts with "IO safety violation" if it encounters an already closed FD, as Rust-the-language declares that there should be a 1:1 correspondence between FDs and `File` objects.

src/event_notifier.rs
```diff
@@ -0,0 +1,291 @@
...
+                self.wfd.as_raw_fd(),
+                &v as *const u64 as *const c_void,
+                mem::size_of::<u64>(),
+            )
+        };
+        if ret <= 0 {
+            Err(io::Error::last_os_error())
+        } else {
+            Ok(())
+        }
```
roypat, line 150: I realize this is copied from eventfd.rs, but we can use the standard library for this:
```suggestion
	self.wfd.write_all(&v.to_ne_bytes())
```
So if you end up respinning this, could you also update eventfd.rs? :o

src/event_notifier.rs
```diff
@@ -0,0 +1,291 @@
...
+                self.rfd.as_raw_fd(),
+                &mut buf as *mut u64 as *mut c_void,
+                mem::size_of::<u64>(),
+            )
+        };
+        if ret < 0 {
+            Err(io::Error::last_os_error())
+        } else {
+            Ok(buf)
+        }
```
roypat, line 191: Same here:
```suggestion
	let mut buf = [0u8; std::mem::size_of::<u64>()];
	self.rfd.read_exact(buf.as_mut_slice())
		.map(|_| u64::from_ne_bytes(buf))
```

src/event_notifier.rs
```diff
@@ -0,0 +1,291 @@
...
+        if rfd < 0 || wfd < 0 {
+            Err(io::Error::last_os_error())
+        } else {
+            Ok(EventNotifier {
+                // SAFETY: This is safe because we checked ret for success and know the kernel
+                // gave us an fd that we own.
+                rfd: unsafe { File::from_raw_fd(rfd) },
+                wfd: unsafe { File::from_raw_fd(wfd) },
+            })
+        }
```
roypat, line 229: and here:
```suggestion
        Ok(EventNotifier {
        	rfd: self.rfd.try_clone()?,
        	wfd: self.wfd.try_clone()?
        })
```

src/event_notifier.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new() {
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        EventNotifier::new(EFD_NONBLOCK).unwrap();
```
roypat, line 240: Maybe we'll want to create an enum containing the flags that are common between Mac and linux, which then internally get translated into the platform-specific flags in some sort of utility constructor, so that callers that want to be agnostic don't have to do with cfg-ery? (but keep a version of the constructor that takes a raw `flags` argument, as we dont want to restrict usage of this library to flags that are common between mac and linux)

roypat (changes requested):
Thanks for your PR!

Since linux and mac support in the same binary is mutually exclusive, I think a cleaner approach here might be to do something similar to the Rust stdlib's "platform abstraction layer". E.g. have two files, one of them can be `linux/eventfd.rs`, the other maybe `mac/event_notifier.rs` that provide structs with the same name and the same impls in both of them, and then a top-level `eventfd.rs` that reexports the correct platform-specific module. 

I think we can even test the macos implementation on linux (since pipes exist on linux), by simply having the macos/event_notifier.rs module be private, but not compiled out.

stefano-garzarella (commented):
Great work! I left some comments.

src/unix/pipefd.rs
```diff
@@ -0,0 +1,166 @@
...
+
+pub use libc::{O_CLOEXEC, O_NONBLOCK};
+
+/// A safe wrapper around Unix
+/// [`pipe`](https://man7.org/linux/man-pages/man2/pipe.2.html).
+#[derive(Debug)]
+pub struct PipeFd {
+    rfd: File,
+    wfd: File,
+}
```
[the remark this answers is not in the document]
stefano-garzarella: > > @roypat do you think we can bump our rust in the CI container to 1.87?
> > If not, I'd suggest following the abstraction, because we will need something like that IMO in vhost-user, (e.g. create only a PipeWriter from the received from the frontend)`` PipeReader/PipeWriter``fd ``
> 
> Do we need to make an abstraction of pipefd and eventfd? Using PipeReader/PipeWriter seems difficult to make an abstraction of pipefd and eventfd.

Good point. Do you have a PoC on how to use this code in vhost-user devices?
Also a simple application that mimic them, because it is not clear to me how can you use it since the VMM (e.g. QEMU) will send only 1 fd.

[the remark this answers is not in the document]
stefano-garzarella: @uran0sH that is the frontend, our goal is to have the backend working on macOS/BSD.
I agree it's nice to add the support also in the fronted, but out priority is the backend.

In our scenario, the frontend is not using rust-vmm creates. Our frontend is QEMU where the pipe is already implemented. QEMU will send to the backend (rust-vmm code) the fd of one side of the pipe.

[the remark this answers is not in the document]
uran0sH: For the backend, if we use std::io::pipe，I think we can use std::io:pipe directly in vhost-user-backend, since std::io::pipe has already implemented the read() for PipeReader and write() for PipeWriter.

uran0sH:
> @uran0sH I just want to add to remember we already discussed the API of this, I don't think it's ok to have `write()` and `read()` as part of the API of a pipe endpoint. And since you are introducing a new type `EventNotifier` why using the name `write/read` and not something more specific.

I just refered the Qemu's implementation of this. https://gitlab.com/qemu-project/qemu/-/blob/master/util/event_notifier-posix.c?ref_type=heads

PR #244, opened 2025-07-09
Implement EventNotifier/EventConsumer as a generic event notification

uran0sH:
### Summary of the PR

Introduce an abstract event notification in order to support multiple platforms. EventNotifier is used to send a notification, and EventConsumer is used to receive a notification. The fd used in EventNotifier or EventConsumer can be from eventfd, pipefd or others.

They can be used like this: https://github.com/uran0sH/vhost/pull/1
### Requirements

requirements:

- [ x ] All commits in this PR have Signed-Off-By trailers (with
- [ x ] All added/changed functionality has a corresponding unit/integration
  test.
- [ x ] All added/changed public-facing functionality has entries in the "Upcoming 
- [ x ] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 -[[#245](https://github.com/rust-vmm/vmm-sys-util/pull/245)]: Make sock_ctrl_msg support unix.
 
+### Added
+
+- [[#244](https://github.com/rust-vmm/vmm-sys-util/pull/244)]: 
+  - Impl `IntoRawFd` for `linux::eventfd::EventFd`.
+  - Use `File::try_clone` to replace the original implementation of `EventFd::try_clone`.
+  Some flags can now be propagated, like `CLOEXEC`.
+  - Add `EventNotifier` and `EventConsumer` as a generic event notification 
+
+
 ## v0.14.0
 
 ### Changed
--- a/src/linux/eventfd.rs
+++ b/src/linux/eventfd.rs
@@ -8,6 +8,7 @@
 //! [`eventfd`](http://man7.org/linux/man-pages/man2/eventfd.2.html).
 
 use std::fs::File;
+use std::os::fd::IntoRawFd;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::{io, mem, result};
 
@@ -167,6 +168,12 @@ impl FromRawFd for EventFd {
         EventFd {
             eventfd: File::from_raw_fd(fd),
         }
+    }
+}
+
+impl IntoRawFd for EventFd {
+    fn into_raw_fd(self) -> RawFd {
+        self.eventfd.into_raw_fd()
     }
 }
 
--- a/src/linux/eventfd.rs
+++ b/src/linux/eventfd.rs
@@ -12,7 +12,7 @@ use std::os::fd::IntoRawFd;
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::{io, mem, result};
 
-use libc::{c_void, dup, eventfd, read, write};
+use libc::{c_void, eventfd, read, write};
 
 // Reexport commonly used flags from libc.
 pub use libc::{EFD_CLOEXEC, EFD_NONBLOCK, EFD_SEMAPHORE};
@@ -142,18 +142,9 @@ impl EventFd {
     /// assert_eq!(evt_clone.read().unwrap(), 923);
     /// ```
     pub fn try_clone(&self) -> result::Result<EventFd, io::Error> {
-        // SAFETY: This is safe because we made this fd and properly check that it returns
-        // without error.
-        let ret = unsafe { dup(self.as_raw_fd()) };
-        if ret < 0 {
-            Err(io::Error::last_os_error())
-        } else {
-            Ok(EventFd {
-                // SAFETY: This is safe because we checked ret for success and know the kernel
-                // gave us an fd that we own.
-                eventfd: unsafe { File::from_raw_fd(ret) },
-            })
-        }
+        Ok(EventFd {
+            eventfd: self.eventfd.try_clone()?,
+        })
     }
 }
 
--- a/src/unix/event.rs
+++ b/src/unix/event.rs
@@ -1,0 +1,316 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//! This module provides a platform-independent interface for event notification,
+//! enabling support across multiple operating systems.
+//!
+//! Internally, it uses a file descriptor that can be backed by various mechanisms
+//! such as `eventfd`, `pipe`, or other OS-specific primitives.
+
+use std::io::{Read, Write};
+use std::os::fd::RawFd;
+use std::{
+    fs::File,
+    io,
+    os::fd::{AsRawFd, FromRawFd, IntoRawFd},
+    result::Result,
+};
+
+bitflags::bitflags! {
+    /// EventFlag
+    /// This enum is used to define flags for the event notifier and consumer.
+    pub struct EventFlag: u8 {
+        /// Non-blocking flag
+        const NONBLOCK = 1 << 0;
+        /// Close-on-exec flag
+        const CLOEXEC = 1 << 1;
+    }
+}
+
+/// EventNotifier
+/// This is a generic event notifier that can be used with eventfd or pipefd.
+/// It allows writing a value to the file descriptor to notify an event.
+///
+/// # Examples
+///
+/// ```
+/// use std::os::fd::FromRawFd;
+/// use std::os::unix::io::IntoRawFd;
+/// use vmm_sys_util::event::EventNotifier;
+/// let (_, writer) = std::io::pipe().expect("Failed to create pipe");
+/// let notifier = unsafe { EventNotifier::from_raw_fd(writer.into_raw_fd()) };
+/// ```
+#[derive(Debug)]
+pub struct EventNotifier {
+    fd: File,
+}
+
+impl EventNotifier {
+    /// Write a value to the EventNotifier's fd
+    /// Writing 1 to fd is for compatibility with Eventfd
+    pub fn notify(&self) -> Result<(), io::Error> {
+        let v = 1u64;
+        (&self.fd).write_all(&v.to_ne_bytes())
+    }
+
+    /// Clone this EventNotifier.
+    pub fn try_clone(&self) -> Result<EventNotifier, io::Error> {
+        Ok(EventNotifier {
+            fd: self.fd.try_clone()?,
+        })
+    }
+}
+
+impl AsRawFd for EventNotifier {
+    fn as_raw_fd(&self) -> RawFd {
+        self.fd.as_raw_fd()
+    }
+}
+
+impl FromRawFd for EventNotifier {
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        EventNotifier {
+            fd: File::from_raw_fd(fd),
+        }
+    }
+}
+
+impl IntoRawFd for EventNotifier {
+    fn into_raw_fd(self) -> RawFd {
+        self.fd.into_raw_fd()
+    }
+}
+
+/// EventConsumer
+/// This is a generic event consumer that can be used with eventfd or pipefd.
+/// It allows reading a value from the file descriptor to consume an event.
+///
+/// # Examples
+///
+/// ```
+/// use std::os::fd::FromRawFd;
+/// use std::os::unix::io::IntoRawFd;
+/// use vmm_sys_util::event::EventConsumer;
+/// let (reader, _) = std::io::pipe().expect("Failed to create pipe");
+/// let consumer = unsafe { EventConsumer::from_raw_fd(reader.into_raw_fd()) };
+/// ```
+#[derive(Debug)]
+pub struct EventConsumer {
+    fd: File,
+}
+
+impl EventConsumer {
+    /// Read a value from the EventConsumer.
+    pub fn consume(&self) -> Result<(), io::Error> {
+        let mut buf = [0u8; size_of::<u64>()];
+        (&self.fd).read_exact(buf.as_mut_slice()).map(|_| Ok(()))?
+    }
+
+    /// Clone this EventConsumer.
+    pub fn try_clone(&self) -> Result<EventConsumer, io::Error> {
+        Ok(EventConsumer {
+            fd: self.fd.try_clone()?,
+        })
+    }
+}
+
+impl AsRawFd for EventConsumer {
+    fn as_raw_fd(&self) -> RawFd {
+        self.fd.as_raw_fd()
+    }
+}
+
+impl FromRawFd for EventConsumer {
+    unsafe fn from_raw_fd(fd: RawFd) -> Self {
+        EventConsumer {
+            fd: File::from_raw_fd(fd),
+        }
+    }
+}
+
+impl IntoRawFd for EventConsumer {
+    fn into_raw_fd(self) -> RawFd {
+        self.fd.into_raw_fd()
+    }
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+unsafe fn fcntl_setfl(fd: RawFd, flag: i32) -> Result<(), io::Error> {
+    let flags = libc::fcntl(fd, libc::F_GETFL);
+    if flags < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    let ret = libc::fcntl(fd, libc::F_SETFL, flags | flag);
+    if ret < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    Ok(())
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+unsafe fn fcntl_setfd(fd: RawFd, flag: i32) -> Result<(), io::Error> {
+    let flags = libc::fcntl(fd, libc::F_GETFD);
+    if flags < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    let ret = libc::fcntl(fd, libc::F_SETFD, flags | flag);
+    if ret < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    Ok(())
+}
+
+/// Create a new EventNotifier and EventConsumer using a pipe.
+///
+/// # Arguments
+///
+/// * `flags` - Flags to set on the file descriptor, such as `EventFlag::NONBLOCK` or `EventFlag::CLOEXEC`.
+///
+/// # Examples
+///
+/// ```
+/// use vmm_sys_util::event::{new_event_consumer_and_notifier, EventFlag};
+/// let (consumer, notifier) = new_event_consumer_and_notifier(EventFlag::NONBLOCK)
+///     .expect("Failed to create notifier and consumer");
+/// notifier.notify().unwrap();
+/// assert!(consumer.consume().is_ok());
+/// ```
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+pub fn new_event_consumer_and_notifier(
+    flags: EventFlag,
+) -> Result<(EventConsumer, EventNotifier), io::Error> {
+    // Use a pipe for non-Linux platforms.
+    let mut fds: [RawFd; 2] = [-1, -1];
+    // SAFETY: This is safe because pipe merely allocated a read fd and a write fd
+    // for our process and we handle the error case.
+    let ret = unsafe { libc::pipe(fds.as_mut_ptr()) };
+    if ret < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let consumer = unsafe { EventConsumer::from_raw_fd(fds[0]) };
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let notifier = unsafe { EventNotifier::from_raw_fd(fds[1]) };
+    if flags.contains(EventFlag::NONBLOCK) {
+        // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
+        // If an exception is thrown, the fds will be closed.
+        unsafe {
+            fcntl_setfl(consumer.as_raw_fd(), libc::O_NONBLOCK)?;
+            fcntl_setfl(notifier.as_raw_fd(), libc::O_NONBLOCK)?;
+        }
+    }
+    if flags.contains(EventFlag::CLOEXEC) {
+        // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
+        // If an exception is thrown, the fds will be closed.
+        unsafe {
+            fcntl_setfd(consumer.as_raw_fd(), libc::FD_CLOEXEC)?;
+            fcntl_setfd(notifier.as_raw_fd(), libc::FD_CLOEXEC)?;
+        }
+    }
+    Ok((consumer, notifier))
+}
+
+/// Create a new EventNotifier and EventConsumer using eventfd.
+///
+/// # Arguments
+///
+/// * `flags` - Flags to set on the file descriptor, such as `EventFlag::NONBLOCK` or `EventFlag::CLOEXEC`.
+///
+/// # Examples
+///
+/// ```
+/// use vmm_sys_util::event::{new_event_consumer_and_notifier, EventFlag};
+/// let (consumer, notifier) = new_event_consumer_and_notifier(EventFlag::NONBLOCK)
+///     .expect("Failed to create consumer and notifier");
+/// notifier.notify().unwrap();
+/// assert!(consumer.consume().is_ok());
+/// ```
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn new_event_consumer_and_notifier(
+    flags: EventFlag,
+) -> Result<(EventConsumer, EventNotifier), io::Error> {
+    let mut efd_flags = 0;
+    if flags.contains(EventFlag::NONBLOCK) {
+        efd_flags |= libc::EFD_NONBLOCK;
+    }
+    if flags.contains(EventFlag::CLOEXEC) {
+        efd_flags |= libc::EFD_CLOEXEC;
+    }
+    let eventfd = crate::linux::eventfd::EventFd::new(efd_flags)?;
+    let eventfd_clone = eventfd.try_clone()?;
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let consumer = unsafe { EventConsumer::from_raw_fd(eventfd.into_raw_fd()) };
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let notifier = unsafe { EventNotifier::from_raw_fd(eventfd_clone.into_raw_fd()) };
+    Ok((consumer, notifier))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::{io::pipe, os::fd::IntoRawFd};
+
+    #[test]
+    fn test_notify_and_consume() {
+        let (reader, writer) = pipe().expect("Failed to create pipe");
+        // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+        let notifier = unsafe { EventNotifier::from_raw_fd(writer.into_raw_fd()) };
+        // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+        let consumer = unsafe { EventConsumer::from_raw_fd(reader.into_raw_fd()) };
+
+        notifier.notify().unwrap();
+        assert!(consumer.consume().is_ok());
+    }
+
+    #[test]
+    fn test_clone() {
+        let (reader, writer) = pipe().expect("Failed to create pipe");
+        // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+        let notifier = unsafe { EventNotifier::from_raw_fd(writer.into_raw_fd()) };
+        // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+        let consumer = unsafe { EventConsumer::from_raw_fd(reader.into_raw_fd()) };
+
+        let cloned_notifier = notifier.try_clone().expect("Failed to clone notifier");
+        let cloned_consumer = consumer.try_clone().expect("Failed to clone consumer");
+
+        cloned_notifier.notify().unwrap();
+        assert!(cloned_consumer.consume().is_ok());
+    }
+
+    #[test]
+    fn test_new_event_notifier_and_consumer() {
+        let (consumer, notifier) = new_event_consumer_and_notifier(EventFlag::empty())
+            .expect("Failed to create notifier and consumer");
+        notifier.notify().unwrap();
+        assert!(consumer.consume().is_ok());
+    }
+
+    #[test]
+    fn test_nonblock() {
+        let (consumer, _notifier) = new_event_consumer_and_notifier(EventFlag::NONBLOCK)
+            .expect("Failed to create notifier and consumer");
+        let r = consumer.consume();
+        match r {
+            Err(ref inner) if inner.kind() == io::ErrorKind::WouldBlock => (),
+            _ => panic!("Unexpected"),
+        }
+    }
+
+    #[test]
+    fn test_cloexec() {
+        let (consumer, notifier) = new_event_consumer_and_notifier(EventFlag::CLOEXEC)
+            .expect("Failed to create notifier and consumer");
+        // SAFETY: This is safe because we check the fd and the return value are valid.
+        let flags = unsafe { libc::fcntl(consumer.as_raw_fd(), libc::F_GETFD) };
+        assert!(flags >= 0 && flags & libc::FD_CLOEXEC == 1);
+        // SAFETY: This is safe because we check the fd and the return value are valid.
+        let flags = unsafe { libc::fcntl(notifier.as_raw_fd(), libc::F_GETFD) };
+        assert!(flags >= 0 && flags & libc::FD_CLOEXEC == 1);
+        let consumer_cloned = consumer.try_clone().expect("Failed to clone consumer");
+        // SAFETY: This is safe because we check the fd and the return value are valid.
+        let flags = unsafe { libc::fcntl(consumer_cloned.as_raw_fd(), libc::F_GETFD) };
+        assert!(flags >= 0 && flags & libc::FD_CLOEXEC == 1);
+        let notifier_cloned = notifier.try_clone().expect("Failed to clone notifier");
+        // SAFETY: This is safe because we check the fd and the return value are valid.
+        let flags = unsafe { libc::fcntl(notifier_cloned.as_raw_fd(), libc::F_GETFD) };
+        assert!(flags >= 0 && flags & libc::FD_CLOEXEC == 1);
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,6 @@ with-serde = ["serde", "serde_derive"]
 libc = "0.2.127"
 serde = { version = "1.0.27", optional = true }
 serde_derive = { version = "1.0.27", optional = true }
-
-[target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
 bitflags = "1.0"
 
 [dev-dependencies]
--- a/src/unix/mod.rs
+++ b/src/unix/mod.rs
@@ -1,5 +1,6 @@
 // Copyright 2022 rust-vmm Authors or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: BSD-3-Clause
+pub mod event;
 pub mod file_traits;
 pub mod sock_ctrl_msg;
 pub mod tempdir;


uran0sH:
Could we bump the rustc version to v1.87.0? If not, I can use libc::pipe() to replace the std::io::pipe()

stefano-garzarella:
> Could we bump the ci's rust version to v1.87.0? If not, I can use libc::pipe() to replace the std::io::pipe()

Yep, as we discussed here: https://github.com/rust-vmm/vmm-sys-util/pull/242#discussion_r2175265742 it should be fine.
We need to increase the rust version in https://github.com/rust-vmm/rust-vmm-container and then update the container version in https://github.com/rust-vmm/rust-vmm-ci , let me know if you need any help with it.

src/event.rs
```diff
@@ -0,0 +1,119 @@
...
+/// EventNotifier
+/// This is a generic event notifier that can be used with eventfd or pipefd.
+/// It allows writing a value to the file descriptor to notify an event.
+#[derive(Debug)]
+pub struct EventNotifier {
+    fd: File,
+}
+
+impl EventNotifier {
+    /// Write a value to the EventNotifier's fd
```
stefano-garzarella, line 21: This is not writing a generic "value", but a specific `1`, I guess to be compatible with eventfd, right?
I think we should document that.

stefano-garzarella (commented):
Great work! I left small comments, but overall LGTM!

For maintainers, this is the continuation of https://github.com/rust-vmm/vmm-sys-util/pull/242 and it's needed for the GSoC project @uran0sH is doing to support vhost-user devices on macOS/BSD.

stefano-garzarella:
@roypat for our use case (vhost-user backend), we receive the `fd` from QEMU through the unix socket, so we really need the `from_raw_fd`, but I see your point, so if @uran0sH want to implement now, it's fine, or we can postpone when we need it.

src/event.rs
roypat, line 1: this file should live in the `unix` module because it unconditionally makes use of items from `std::os::unix`

roypat:
> @roypat for our use case (vhost-user backend), we receive the `fd` from QEMU through the unix socket, so we really need the `from_raw_fd`, but I see your point, so if @uran0sH want to implement now, it's fine, or we can postpone when we need it.

Ah, yeah, I didn't mean we shouldn't have the from_raw_fd APIs, sorry. I think either is fine by me, but if we don't create some sort of `pipe()`-like constructor, then let's add something about how `FromRawFd` must be used for construction to the API docs, because it took me a minute of staring at the code to understand how one is supposed to construct these haha

src/event.rs
```diff
@@ -0,0 +1,119 @@
...
+/// This is a generic event consumer that can be used with eventfd or pipefd.
+/// It allows reading a value from the file descriptor to consume an event.
+#[derive(Debug)]
+pub struct EventConsumer {
+    fd: File,
+}
+
+impl EventConsumer {
+    /// Read a value from the EventConsumer.
+    pub fn consume(&self) -> result::Result<u64, io::Error> {
```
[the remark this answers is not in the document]
germag: I agree, 
I think this is related with the `notify()` so:
1- we should allow the user to define the value to be sent with the notification in `notify()`, or
2- `consume()` return type should be `result::Result<(), io::Error>` to be the "lowest common denominator" between pipe() and eventfd

Or... having two methods?:
- like `notify()` that always sends a `1`, and `pub fn consume(&self) -> result::Result<(), io::Error>`
- and `notify_with_value(v: u64)`, and `pub fn consume_with_value(&self) -> result::Result<u64, io::Error>` (the names are terrible, I know)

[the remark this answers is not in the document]
stefano-garzarella: yep, IMO we should do your option 2 and for now avoid `notify_with_value(v: u64)`, if we will need it (not sure for what use case), we can add later.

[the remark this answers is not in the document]
germag: Agreed

uran0sH:
> > Ah, yeah, I didn't mean we shouldn't have the from_raw_fd APIs, sorry. I think either now or later is fine by me, but if we don't create some sort of `pipe()`-like constructor, then let's add something about how `FromRawFd` must be used for construction to the API docs, because it took me a minute of staring at the code to understand how one is supposed to construct these haha
> 
> Yeah, +1 on this! @uran0sH ^

Add an example here: https://github.com/rust-vmm/vmm-sys-util/pull/244/files#diff-b817124a7a414d687ff2826960adbb7bf304a2770b0f6ee70b3daa5ca4a16113R46

roypat:
> > > Ah, yeah, I didn't mean we shouldn't have the from_raw_fd APIs, sorry. I think either now or later is fine by me, but if we don't create some sort of `pipe()`-like constructor, then let's add something about how `FromRawFd` must be used for construction to the API docs, because it took me a minute of staring at the code to understand how one is supposed to construct these haha
> > 
> > 
> > Yeah, +1 on this! @uran0sH ^
> 
> Add an example here: [#244 (files)](https://github.com/rust-vmm/vmm-sys-util/pull/244/files#diff-b817124a7a414d687ff2826960adbb7bf304a2770b0f6ee70b3daa5ca4a16113R46)

thanks! could you move the example to the struct definition though? Having it on the trait impl itself doesn't really help unless someone already knows they should be looking for a `FromRawFd` impl :p

uran0sH:
> > > > Ah, yeah, I didn't mean we shouldn't have the from_raw_fd APIs, sorry. I think either now or later is fine by me, but if we don't create some sort of `pipe()`-like constructor, then let's add something about how `FromRawFd` must be used for construction to the API docs, because it took me a minute of staring at the code to understand how one is supposed to construct these haha
> > > 
> > > 
> > > Yeah, +1 on this! @uran0sH ^
> > 
> > 
> > Add an example here: [#244 (files)](https://github.com/rust-vmm/vmm-sys-util/pull/244/files#diff-b817124a7a414d687ff2826960adbb7bf304a2770b0f6ee70b3daa5ca4a16113R46)
> 
> thanks! could you move the example to the struct definition though? Having it on the trait impl itself doesn't really help unless someone already knows they should be looking for a `FromRawFd` impl :p

done

stefano-garzarella:
@uran0sH what about https://github.com/rust-vmm/vmm-sys-util/pull/244#pullrequestreview-3002187055 ?

Can we re-add some cleanups and an API to construct (EventNotifier, EventConsumer) automatically with eventfd() or fallback to pipe(). In this way we can use it in https://github.com/rust-vmm/vhost/pull/308#discussion_r2195469019

src/unix/event.rs
```diff
@@ -0,0 +1,183 @@
...
+        let notifier = unsafe { EventNotifier::from_raw_fd(writer.into_raw_fd()) };
+        // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+        let consumer = unsafe { EventConsumer::from_raw_fd(reader.into_raw_fd()) };
+        Ok((notifier, consumer))
+    }
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    {
+        // SAFETY: This is safe because eventfd merely allocated an eventfd for
+        // our process and we handle the error case.
+        let fd = unsafe { libc::eventfd(0, libc::EFD_NONBLOCK) };
```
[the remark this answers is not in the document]
stefano-garzarella: done, what?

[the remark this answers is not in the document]
uran0sH: > Why it was non-blocking and now it's blocking? Why not using `linux::event_fd` as Patrick suggested?
1. Keep the same as pipe(). std::io::pipe is blocking.
2. 
`linux::event_fd` doens't provide `into_raw_fd()`, only provides `as_raw_fd`. This will cause `IO Safety violation: owned file descriptor already closed, aborting`. We can do it like this:
```rust
    let evt = linux::eventfd::EventFd::new(0).unwrap();
    let evt_clone = evt.try_clone()?;
    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
    let notifier = unsafe { EventNotifier::from_raw_fd(evt.as_raw_fd()) };
    
    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
    let consumer = unsafe { EventConsumer::from_raw_fd(evt_clone.as_raw_fd()) };
    std::mem::forget(evt);
    std::mem::forget(evt_clone);
```
Using `libc::eventfd` can avoid using `std::mem:forget()`. And I thought we only need fd, not other encapsulated methods.

src/linux/eventfd.rs
```diff
@@ -170,6 +171,12 @@ impl FromRawFd for EventFd {
     }
 }
 
+impl IntoRawFd for EventFd {
```
stefano-garzarella, line 174: Please, move this in a separate commit.

src/unix/event.rs
```diff
@@ -0,0 +1,186 @@
...
+        EventConsumer {
+            fd: File::from_raw_fd(fd),
+        }
+    }
+}
+
+/// Create a new EventNotifier and EventConsumer using a pipe.
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+pub fn new_event_notifier_and_consumer(
+    flag: i32,
```
stefano-garzarella, line 112: We should document what flag is supposed to be.
Are we sure it's the same for `fcntl` and for `EventFd::new()` ?

src/linux/eventfd.rs
roypat: Potentially could also get a CHANGELOG entry

src/unix/event.rs
roypat: Yeah, something like that! Although like this it only allows to set one flag at a time, instead of a combination of them (sorry, been working in C too much recently that I apparently forgot how Rust enums work :sob:), so we'll need something like [`bitflags!`](https://docs.rs/bitflags/latest/bitflags/) (to clarify, the value of each flag should be a synthetic value introduced by this crate, e.g. NonBlock = 0x1, CloseOnExec = 0x2, and then `From::from` checks which synthetic bits are set and converts it to the platform specific combination of bits that correspond to them)

germag: Sorry if I'm late here, but I prefer a builder type of API, that can handle both O_CLOEXEC and NONBLOCK since both are common to pipes and eventfd, something like:
```rust
let (en, ec) = EventBuilder::new()
    .set_cloexec()
    .set_nonblock()
    .build()
```

so the flags are optional


for instance we do care about cloexec in virtiofsd

germag: I'm not suggesting the names btw, we can also do something similar to what is done inside the std
```rust
let (en, ec) = EventBuilder::new()
    .closeOnExec(true)
    .nonBlock(true)
    .build()
```

or without the bool parameter

src/linux/eventfd.rs
stefano-garzarella: Not for each commit, but for each user visible change. It's fine if the link to the same PR.

src/unix/event.rs
```diff
@@ -0,0 +1,254 @@
...
+            fd: File::from_raw_fd(fd),
+        }
+    }
+}
+
+/// Create a new EventNotifier and EventConsumer using a pipe.
+///
+/// # Arguments
+///
+/// * `flag` - Flags to set on the file descriptor, such as `O_NONBLOCK` or `O_CLOEXEC`.
```
stefano-garzarella, line 152: `Flags to set ` I don't understand how can we pass multiple flags...

src/unix/event.rs
```diff
@@ -0,0 +1,254 @@
...
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let consumer = unsafe { EventConsumer::from_raw_fd(reader.into_raw_fd()) };
+    Ok((notifier, consumer))
+}
+
+/// Create a new EventNotifier and EventConsumer using eventfd.
+///
+/// # Arguments
+///
+/// * `flag` - Flags to set on the eventfd, such as `EFD_NONBLOCK`, `EFD_CLOEXEC` or `EFD_SEMAPHORE`.
```
stefano-garzarella, line 191: Here we should document the `EventFlag` options, not the underlying mapping

src/unix/event.rs
```diff
@@ -0,0 +1,254 @@
...
+    fs::File,
+    io,
+    os::fd::{AsRawFd, FromRawFd, IntoRawFd},
+    result,
+};
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub use libc::{EFD_CLOEXEC, EFD_NONBLOCK};
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+pub use libc::{O_CLOEXEC, O_NONBLOCK};
```
stefano-garzarella, line 15: Why they are `pub` ?

src/unix/event.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+impl From<EventFlag> for i32 {
+    fn from(flag: EventFlag) -> Self {
+        match flag {
+            EventFlag::NoFlag => 0,
+            EventFlag::NonBlock => libc::O_NONBLOCK,
+            EventFlag::CloseOnExec => libc::O_CLOEXEC,
```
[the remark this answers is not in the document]
uran0sH: Yeah, it can't work. `fcntl` need to use `FD_CLOEXEC`

src/unix/event.rs
uran0sH: Yes, it cannot pass multiple flags. I will modify it to use bitflags later.

src/unix/event.rs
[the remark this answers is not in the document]
roypat: I have a slight preference for bitflags tbh, as they'll need a less boilerplate, both in this crate and for users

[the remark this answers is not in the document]
stefano-garzarella: I re-read the @germag and it seems virtiofsd needs also cloexec stuff, so +1 about builder type API

src/unix/event.rs
germag: it's not required, just expose the bitflags

src/unix/event.rs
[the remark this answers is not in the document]
stefano-garzarella: @roypat okay, so let's go to the bitflags, but taking care of O_CLOEXEC in the right way.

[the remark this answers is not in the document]
uran0sH: The use of bitflags is now implemented.

roypat (approved):

src/unix/event.rs
```diff
@@ -0,0 +1,265 @@
...
+    // Use a pipe for non-Linux platforms.
+    let (reader, writer) = std::io::pipe()?;
+    // SAFETY: Safe because we check the fd is valid and check the return value.
+    let flags = unsafe { libc::fcntl(reader.as_raw_fd(), libc::F_GETFL) };
+    if flags < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    let flag: i32 = flag.into();
+    // SAFETY: Safe because we check the fd is valid and check the return value.
+    let ret = unsafe { libc::fcntl(reader.as_raw_fd(), libc::F_SETFL, flags | flag) };
```
stefano-garzarella, line 176: From `man 2 fcntl`:

```
   File descriptor flags
       The  following  operations  manipulate the flags associated with a file descriptor.  Currently, only one such flag is defined: FD_CLOEXEC, the close-on-exec flag.  If
       the FD_CLOEXEC bit is set, the file descriptor will automatically be closed during a successful execve(2).  (If the execve(2)  fails,  the  file  descriptor  is  left
       open.)  If the FD_CLOEXEC bit is not set, the file descriptor will remain open across an execve(2).

       F_GETFD (void)
              Return (as the function result) the file descriptor flags; arg is ignored.

       F_SETFD (int)
              Set the file descriptor flags to the value specified by arg.

       In multithreaded programs, using fcntl() F_SETFD to set the close-on-exec flag at the same time as another thread performs a fork(2) plus execve(2) is vulnerable to a
       race  condition  that  may unintentionally leak the file descriptor to the program executed in the child process.  See the discussion of the O_CLOEXEC flag in open(2)
       for details and a remedy to the problem.
```

So IIUC FD_CLOEXEC should not be used with F_SETFL but with F_SETFD.
Did you test this code?
Should we add a test for it?

uran0sH: After checking, it should use F_SETFD. We should add a unit test to cover it.

stefano-garzarella: Yep, so IIUC we should use F_SETFL for O_* flags and F_SETFD for FD__* flags, rigth?

uran0sH: > Yep, so IIUC we should use F_SETFL for O_* flags and F_SETFD for FD__* flags, rigth?
yeah

uran0sH: I found that `std::io::pipe` will set the created fd to `CLOEXEC` by default(https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pal/unix/pipe.rs#L43). So in order to keep the new_event_notifier_and_consumer operation consistent, should we set CLOEXEC to EventFd by default? And then provide a flag to clear CLOEXEC

stefano-garzarella: > I found that `std::io::pipe` will set the created fd to `CLOEXEC` by default(https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pal/unix/pipe.rs#L43). So in order to keep the new_event_notifier_and_consumer operation consistent, should we set CLOEXEC to EventFd by default? And then provide a flag to clear CLOEXEC

@uran0sH oh, interesting. @roypat @germag should we just set it by default ignoring for now the case to clear it? (not sure what is the use case).

uran0sH: I prefer to fallback to use `libc::pipe` and use builder api. Since if we use pipe, we need to use different type of fcntl, like `F_GETFL` and `F_SETFL` for different flags, use builder api might be a better implementation.
```rust
#[derive(Debug)]
pub struct EventBuilder {
    consumer: EventConsumer,
    notifier: EventNotifier,
}

impl EventBuilder {
    ///...
    pub fn new() -> Self {
        let mut fds: [RawFd; 2] = [-1, -1];
        let ret = unsafe {libc::pipe(fds.as_mut_ptr())};
        unsafe {
            Self {
                consumer: EventConsumer::from_raw_fd(fds[0]),
                notifier: EventNotifier::from_raw_fd(fds[1]),
            }
        }
    }

    ///...
    pub fn set_nonblock(self) -> Self {
        let flags = unsafe { libc::fcntl(self.consumer.as_raw_fd(), libc::F_GETFL) };
        // SAFETY: Safe because we check the fd is valid and check the return value.
        let ret = unsafe { libc::fcntl(self.consumer.as_raw_fd(), libc::F_SETFL, flags | libc::O_NONBLOCK) };
        let flags = unsafe { libc::fcntl(self.notifier.as_raw_fd(), libc::F_GETFL) };
        // SAFETY: Safe because we check the fd is valid and check the return value.
        let ret = unsafe { libc::fcntl(self.notifier.as_raw_fd(), libc::F_SETFL, flags | libc::O_NONBLOCK) };
        self
    }

    ///...
    pub fn set_cloexec(self) -> Self {
        let flags = unsafe { libc::fcntl(self.consumer.as_raw_fd(), libc::F_GETFD) };
        // SAFETY: Safe because we check the fd is valid and check the return value.
        let ret = unsafe { libc::fcntl(self.consumer.as_raw_fd(), libc::F_SETFD, flags | libc::FD_CLOEXEC) };
        let flags = unsafe { libc::fcntl(self.notifier.as_raw_fd(), libc::F_GETFD) };
        // SAFETY: Safe because we check the fd is valid and check the return value.
        let ret = unsafe { libc::fcntl(self.notifier.as_raw_fd(), libc::F_SETFD, flags | libc::FD_CLOEXEC) };
        self
    }

    ///...
    pub fn build(self) -> (EventConsumer, EventNotifier) {
        (self.consumer, self.notifier)
    }
}

```

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+    // Use a pipe for non-Linux platforms.
+    use std::os::fd::RawFd;
+    let mut fds: [RawFd; 2] = [-1, -1];
+    // SAFETY: Safe because we check the fd is valid and check the return value.
+    let ret = unsafe { libc::pipe(fds.as_mut_ptr()) };
+    if ret < 0 {
+        return Err(io::Error::last_os_error());
+    }
+    if flag.contains(EventFlag::NONBLOCK) {
+        fcntl_getfl_and_setfl(fds[0], libc::O_NONBLOCK)?;
```
stefano-garzarella, line 215: If this fail, who close the `fd`s we opened?

stefano-garzarella: Same for next block.

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+        return Err(io::Error::last_os_error());
+    }
+    Ok(())
+}
+
+/// Create a new EventNotifier and EventConsumer using a pipe.
+///
+/// # Arguments
+///
+/// * `flag` - Flags to set on the file descriptor, such as `EventFlag::NONBLOCK` or `EventFlag::CLOEXEC`.
```
stefano-garzarella, line 191: nit: I would suggest renaming `flag` in `flags`, since also in the description we use the plural.

Please check all the places. (`EventFlag` is fine, but the variable can contain multiple flags, so I'd use the plural for the variable name)

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+    if flag.contains(EventFlag::NONBLOCK) {
+        fcntl_getfl_and_setfl(fds[0], libc::O_NONBLOCK)?;
+        fcntl_getfl_and_setfl(fds[1], libc::O_NONBLOCK)?;
+    }
+    if flag.contains(EventFlag::CLOEXEC) {
+        use libc::FD_CLOEXEC;
+
+        fcntl_getfd_and_setfd(fds[0], FD_CLOEXEC)?;
+        fcntl_getfd_and_setfd(fds[1], FD_CLOEXEC)?;
+    }
```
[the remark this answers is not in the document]
germag: I prefer the `libc::...` version too

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+/// ```
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn new_event_consumer_and_notifier(
+    flag: EventFlag,
+) -> std::result::Result<(EventConsumer, EventNotifier), io::Error> {
+    let eventfd = crate::linux::eventfd::EventFd::new(flag.into())?;
+    let eventfd_clone = eventfd.try_clone()?;
+    if flag.contains(EventFlag::CLOEXEC) {
+        fcntl_getfd_and_setfd(eventfd_clone.as_raw_fd(), libc::FD_CLOEXEC)?;
+    }
```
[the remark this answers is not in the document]
germag: maybe we can change the `EventFd::try_clone()` to something like:
```rust
    pub fn try_clone(&self) -> result::Result<EventFd, io::Error> {
            let eventfd = self.eventfd.try_clone().map_err(...);
            Ok(EventFd {  eventfd })
        }
    }
```
since `File::try_clone()` already takes care of the `FD_CLOEXEC` flag

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+                let path = CString::new("/bin/sleep").expect("CString::new failed");
+                let arg0 = CString::new("sleep").expect("CString::new failed");
+                let arg1 = CString::new("1").expect("CString::new failed");
+                let ret = libc::execl(
+                    path.as_ptr(),
+                    arg0.as_ptr(),
+                    arg1.as_ptr(),
+                    std::ptr::null::<i8>(),
+                );
+                assert_eq!(ret, 0, "execl failed");
```
stefano-garzarella, line 332: Can you explain a bit this test, for example, why you need to call `/bin/sleep` ?

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+    #[test]
+    fn test_new_event_notifier_and_consumer() {
+        let (consumer, notifier) = new_event_consumer_and_notifier(EventFlag::empty())
+            .expect("Failed to create notifier and consumer");
+        notifier.notify().unwrap();
+        assert!(consumer.consume().is_ok());
+    }
+
+    #[test]
+    fn test_read_nothing() {
```
stefano-garzarella, line 304:
```suggestion
    fn test_nonblock() {
```

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+impl IntoRawFd for EventNotifier {
+    fn into_raw_fd(self) -> std::os::unix::prelude::RawFd {
+        self.fd.into_raw_fd()
+    }
+}
+
+/// EventReceiver
```
stefano-garzarella, line 105:
```suggestion
/// EventConsumer
```

src/unix/event.rs
stefano-garzarella: okay, IIUC it's just something to pass to `execl`, but how we are sure the child called `execl` before the parent calls `lsof` ?

src/unix/event.rs
```diff
@@ -0,0 +1,346 @@
...
+
+impl IntoRawFd for EventConsumer {
+    fn into_raw_fd(self) -> std::os::unix::prelude::RawFd {
+        self.fd.into_raw_fd()
+    }
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+fn fcntl_getfl_and_setfl(fd: i32, flag: i32) -> std::result::Result<(), io::Error> {
+    // SAFETY: Safe because we check the fd is valid and check the return value.
```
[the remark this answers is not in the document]
uran0sH: Change to unsafe and check the fd in new_event_consumer_and_notifier

src/unix/event.rs
uran0sH: https://github.com/rust-vmm/vmm-sys-util/pull/244/files#diff-b817124a7a414d687ff2826960adbb7bf304a2770b0f6ee70b3daa5ca4a16113R329 Make sure the process exec sleep through monitoring the cmdline of the process

src/unix/event.rs
uran0sH: Add these lines (https://github.com/rust-vmm/vmm-sys-util/pull/244/files#diff-b817124a7a414d687ff2826960adbb7bf304a2770b0f6ee70b3daa5ca4a16113R188) to close fd. In another PR, perhaps we can provide a pipe wrapper to better use Rust's RAII

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 -[[#245](https://github.com/rust-vmm/vmm-sys-util/pull/245)]: Make sock_ctrl_msg support unix.
 
+### Added
+
+- [[#244](https://github.com/rust-vmm/vmm-sys-util/pull/244)]: 
+  - Impl `IntoRawFd` for `linux::eventfd::EventFd`.
+  - Use File::try_clone to replace the original implementation of EventFd::try_clone
```
stefano-garzarella, line 11: Please mention the change for the user (e.g. CLOEXEC is now propagated)

src/unix/event.rs
```diff
@@ -0,0 +1,354 @@
...
+            fcntl_setfl(fds[0], libc::O_NONBLOCK).map_err(|e| {
+                libc::close(fds[0]);
+                libc::close(fds[1]);
+                e
+            })?;
+            fcntl_setfl(fds[1], libc::O_NONBLOCK).map_err(|e| {
+                libc::close(fds[0]);
+                libc::close(fds[1]);
+                e
+            })?;
```
stefano-garzarella, line 197: mmm, I don't like this. Can we just use `OwnedFd` or `File`?

I mean something like this:
```
    let ret = unsafe { libc::pipe(fds.as_mut_ptr()) };
    if ret < 0 {
        return Err(io::Error::last_os_error());
    }

    let consumer_fd = unsafe { OwnedFd::from_raw_fd(fds[0]) };
    let producer_fd = unsafe { OwnedFd::from_raw_fd(fds[1]) };

    ...
    fcntl_setfl(consumer_fd.as_raw_fd(), libc::O_NONBLOCK)?;
    ...

    let consumer = unsafe { EventConsumer::from_raw_fd(consumer_fd.into_raw_fd()) };
```

Or even better, just create `EventConsumer` and `EventNotifier` after pipe():
```diff
@@ -181,42 +183,28 @@ pub fn new_event_consumer_and_notifier(
     if ret < 0 {
         return Err(io::Error::last_os_error());
     }
+
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let consumer = unsafe { EventConsumer::from_raw_fd(fds[0]) };
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let notifier = unsafe { EventNotifier::from_raw_fd(fds[1]) };
+
     if flags.contains(EventFlag::NONBLOCK) {
         // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
         // If an exception is thrown, the fds will be closed.
         unsafe {
-            fcntl_setfl(fds[0], libc::O_NONBLOCK).map_err(|e| {
-                libc::close(fds[0]);
-                libc::close(fds[1]);
-                e
-            })?;
-            fcntl_setfl(fds[1], libc::O_NONBLOCK).map_err(|e| {
-                libc::close(fds[0]);
-                libc::close(fds[1]);
-                e
-            })?;
+            fcntl_setfl(consumer.as_raw_fd(), libc::O_NONBLOCK)?;
+            fcntl_setfl(notifier.as_raw_fd(), libc::O_NONBLOCK)?;
         }
     }
     if flags.contains(EventFlag::CLOEXEC) {
         // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
         // If an exception is thrown, the fds will be closed.
         unsafe {
-            fcntl_setfd(fds[0], libc::FD_CLOEXEC).map_err(|e| {
-                libc::close(fds[0]);
-                libc::close(fds[1]);
-                e
-            })?;
-            fcntl_setfd(fds[1], libc::FD_CLOEXEC).map_err(|e| {
-                libc::close(fds[0]);
-                libc::close(fds[1]);
-                e
-            })?;
+            fcntl_setfd(consumer.as_raw_fd(), libc::FD_CLOEXEC)?;
+            fcntl_setfd(notifier.as_raw_fd(), libc::FD_CLOEXEC)?;
         }
     }
-    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
-    let consumer = unsafe { EventConsumer::from_raw_fd(fds[0].into_raw_fd()) };
-    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
-    let notifier = unsafe { EventNotifier::from_raw_fd(fds[1].into_raw_fd()) };
     Ok((consumer, notifier))
 }
```

src/unix/event.rs
```diff
@@ -0,0 +1,354 @@
...
+                }
+                let output = Command::new("lsof")
+                    .args(["-p", pid.to_string().as_str()])
+                    .output()
+                    .expect("Failed to execute lsof command");
+                output.stdout.lines().for_each(|line| {
+                    let line = line.expect("Failed to read line");
+                    assert!(!line.contains("PIPE") && !line.contains("eventfd"));
+                });
+            }
```
stefano-garzarella, line 351: As I mentioned, IMO this is racy also with the new loop.

My suggestion is to have a loop, with the `lsof` inside and try it for a while (e.g. 5 sec like the sleep), I mean something like this in pseudo-code:
```
    let start = Instant::now();
    let timeout = Duration::new(5, 0); // 5 seconds
    let mut passed = false;

    while start.elapsed() < timeout && !passed {
        std::thread::sleep(Duration::from_millis(10));
        
        let output = ...
        ...
           let line = ...
           if line.contain... && ... {
                 passed = true;
           }
     }

     assert!(passed);
```

src/unix/event.rs
stefano-garzarella: Or just create `EventConsumer` and `EventNotifier` after pipe() as I mentioned in https://github.com/rust-vmm/vmm-sys-util/pull/244#discussion_r2238823427

src/unix/event.rs
germag: > @germag I thought `lsof` usage was for supporting the test on any `unix`, but honestly I'm not sure if it's available also on BSD/macOS.

true!  

but why not just call `libc::fcntl(..., F_GETFD)` and check if returns `FD_CLOEXEC`?, I mean we don't need to test if the operating system honor `FD_CLOEXEC` we just need to test if the flag is set

stefano-garzarella: > > @germag I thought `lsof` usage was for supporting the test on any `unix`, but honestly I'm not sure if it's available also on BSD/macOS.
> 
> true!
> 
> but why not just call `libc::fcntl(..., F_GETFD)` and check if returns `FD_CLOEXEC`?, I mean we don't need to test if the operating system honors `FD_CLOEXEC` we just need to test if the flag is set

This is a great point, @uran0sH WDYT?

stefano-garzarella: @uran0sH yes of course, I was talking about the last proposal from @germag:

> but why not just call libc::fcntl(..., F_GETFD) and check if returns FD_CLOEXEC?, I mean we don't need to test if the operating system honors FD_CLOEXEC we just need to test if the flag is set

uran0sH: OK, maybe we can just check if FD_CLOEXEC is set, which would simplify this test

src/unix/event.rs
```diff
@@ -0,0 +1,305 @@
...
+            .expect("Failed to create notifier and consumer");
+        let r = consumer.consume();
+        match r {
+            Err(ref inner) if inner.kind() == io::ErrorKind::WouldBlock => (),
+            _ => panic!("Unexpected"),
+        }
+    }
+
+    #[test]
+    fn test_cloexec() {
```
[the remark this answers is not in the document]
germag: good idea, yes. I think not keeping the flag in `EventFd:.try_clone()` was an oversight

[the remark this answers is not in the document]
stefano-garzarella: @germag yeah.

@uran0sH please add this check in the test and I think we are fine! Great job!

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
 
 -[[#245](https://github.com/rust-vmm/vmm-sys-util/pull/245)]: Make sock_ctrl_msg support unix.
 
+### Added
+
+- [[#244](https://github.com/rust-vmm/vmm-sys-util/pull/244)]: 
+  - Impl `IntoRawFd` for `linux::eventfd::EventFd`.
+  - Use File::try_clone to replace the original implementation of EventFd::try_clone.
+  Some flags can now be propagated, like CLOEXEC.
+  - Add EventNotifier and EventConsumer as a generic event notification
```
germag, line 13: nit: since you used `` above, why not do the same for `File::try_clone()`, `EventFd::try_clone()`, `CLOEXEC`, `EventNotifier` and `EventConsumer` that are also types, methods, etc.

src/unix/event.rs
```diff
@@ -0,0 +1,305 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//! This is an abstract EventNotifer and EventConsumer for eventfd and pipefd
```
stefano-garzarella, line 2:
```suggestion
//! This module provides a platform-independent interface for event notification,
//! enabling support across multiple operating systems.
//!
//! Internally, it uses a file descriptor that can be backed by various mechanisms
//! such as `eventfd`, `pipe`, or other OS-specific primitives.
```

stefano-garzarella (approved):

src/linux/eventfd.rs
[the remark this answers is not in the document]
germag: do you mean replacing `EventFd::write()` and `EventFd::read()`?

https://github.com/rust-vmm/vmm-sys-util/pull/242/files#diff-47e48116fffb91d140d43423133da3239c241f899495f64c8eacf3ea597ca074

@uran0sH it's not that much work wdyt?, if you agree to do it, do it in a separate commit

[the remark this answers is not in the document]
roypat: Yeah, those are the ones I meant. But it's just a nit, so if you dont want to do it in this PR, I'll just submit a follow up after we merge this :)

src/unix/event.rs
```diff
@@ -0,0 +1,316 @@
...
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let consumer = unsafe { EventConsumer::from_raw_fd(fds[0]) };
+    // SAFETY: Safe because we check the fd is valid. And the kernel gave us an fd that we own.
+    let notifier = unsafe { EventNotifier::from_raw_fd(fds[1]) };
+    if flags.contains(EventFlag::NONBLOCK) {
+        // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
+        // If an exception is thrown, the fds will be closed.
+        unsafe {
+            fcntl_setfl(consumer.as_raw_fd(), libc::O_NONBLOCK)?;
+            fcntl_setfl(notifier.as_raw_fd(), libc::O_NONBLOCK)?;
```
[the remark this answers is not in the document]
stefano-garzarella: @uran0sH all structs already implements `AsRawFd`, so can we do something like this:
```
diff --git a/src/unix/event.rs b/src/unix/event.rs
index 8371d7a..6ad1530 100644
--- a/src/unix/event.rs
+++ b/src/unix/event.rs
@@ -133,12 +133,16 @@ impl IntoRawFd for EventConsumer {
 }
 
 #[cfg(not(any(target_os = "linux", target_os = "android")))]
-unsafe fn fcntl_setfl(fd: RawFd, flag: i32) -> Result<(), io::Error> {
-    let flags = libc::fcntl(fd, libc::F_GETFL);
+fn fcntl_setfl(event: &impl AsRawFd, flag: i32) -> Result<(), io::Error> {
+    let fd = event.as_raw_fd();
+    // SAFETY: The file descriptor is guaranteed to be valid by the AsRawFd trait contract.
+    // We check the return value of fcntl to ensure safety.
+    let flags = unsafe { libc::fcntl(fd, libc::F_GETFL) };
     if flags < 0 {
         return Err(io::Error::last_os_error());
     }
-    let ret = libc::fcntl(fd, libc::F_SETFL, flags | flag);
+    // SAFETY: The file descriptor is guaranteed to be valid, and we check the return value.
+    let ret = unsafe { libc::fcntl(fd, libc::F_SETFL, flags | flag) };
     if ret < 0 {
         return Err(io::Error::last_os_error());
     }
@@ -146,12 +150,16 @@ unsafe fn fcntl_setfl(fd: RawFd, flag: i32) -> Result<(), io::Error> {
 }
 
 #[cfg(not(any(target_os = "linux", target_os = "android")))]
-unsafe fn fcntl_setfd(fd: RawFd, flag: i32) -> Result<(), io::Error> {
-    let flags = libc::fcntl(fd, libc::F_GETFD);
+fn fcntl_setfd(event: &impl AsRawFd, flag: i32) -> Result<(), io::Error> {
+    let fd = event.as_raw_fd();
+    // SAFETY: The file descriptor is guaranteed to be valid by the AsRawFd trait contract.
+    // We check the return value of fcntl to ensure safety.
+    let flags = unsafe { libc::fcntl(fd, libc::F_GETFD) };
     if flags < 0 {
         return Err(io::Error::last_os_error());
     }
-    let ret = libc::fcntl(fd, libc::F_SETFD, flags | flag);
+    // SAFETY: The file descriptor is guaranteed to be valid, and we check the return value.
+    let ret = unsafe { libc::fcntl(fd, libc::F_SETFD, flags | flag) };
     if ret < 0 {
         return Err(io::Error::last_os_error());
     }
@@ -192,18 +200,14 @@ pub fn new_event_consumer_and_notifier(
     if flags.contains(EventFlag::NONBLOCK) {
         // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
         // If an exception is thrown, the fds will be closed.
-        unsafe {
-            fcntl_setfl(consumer.as_raw_fd(), libc::O_NONBLOCK)?;
-            fcntl_setfl(notifier.as_raw_fd(), libc::O_NONBLOCK)?;
-        }
+        fcntl_setfl(&consumer, libc::O_NONBLOCK)?;
+        fcntl_setfl(&notifier, libc::O_NONBLOCK)?;
     }
     if flags.contains(EventFlag::CLOEXEC) {
         // SAFETY: We ensure that the passed fd is valid and check the return value of all operations.
         // If an exception is thrown, the fds will be closed.
-        unsafe {
-            fcntl_setfd(consumer.as_raw_fd(), libc::FD_CLOEXEC)?;
-            fcntl_setfd(notifier.as_raw_fd(), libc::FD_CLOEXEC)?;
-        }
+        fcntl_setfd(&consumer, libc::FD_CLOEXEC)?;
+        fcntl_setfd(&notifier, libc::FD_CLOEXEC)?;
     }
     Ok((consumer, notifier))
 }
```

Or use generics if you prefer.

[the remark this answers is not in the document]
roypat: We can't use `AsRawFd`, because we cannot trust its implementation to return a valid fd (it's a safe trait, and `as_raw_fd()` just returns an integer, so someone could do `impl AsRawFd for MyStruct { fn as_raw_fd(&self) -> RawFd { 42 }}`. But we could use [`AsFd`](https://doc.rust-lang.org/std/os/fd/trait.AsFd.html), which returns a `BorrowedFd` instead of an integer, and `BorrowedFd` has an unsafe constructor that requires you to uphold the invariant that the wrapped fd is valid for the livetime associated with the `BorrowedFd`.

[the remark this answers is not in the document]
roypat: Why do we need to implement `AsRef<File>` for what @germag is suggesting? :o (I also don't really think we should give people access to the underlying `File` object, seems a bit leaky of an abstraction to me)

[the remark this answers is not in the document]
stefano-garzarella: What is the difference with `AsRawFd` and `IntoRawFd` we already do?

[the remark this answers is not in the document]
stefano-garzarella: BTW I'd like to have this merged as soon as possible, so I really don't have a strong opinion, we can eventually leave this as it is and fix with a follow-up since it isn't a public API.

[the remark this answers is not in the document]
roypat: Okay, that's fair. I'll post a follow up :)

roypat (approved):

src/linux/eventfd.rs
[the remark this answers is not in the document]
stefano-garzarella: @roypat oooops, sorry, I enabled the auto merge, so after your approval this is merged I guess without fixing this comment. (not sure if we should enabled the option to avoid merges till there are open comment).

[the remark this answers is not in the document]
roypat: ah, no worries, I was aware that would happen :) I'll just do all three comments I had in my follow up

PR #245, opened 2025-07-11
Make sock_ctrl_msg work on macos

uran0sH:
### Summary of the PR

Make sock_ctrl_msg work on macos. Split to these commits:
1. Use pipefd to replace eventfd in tests
2. Replace CMSG_*! macros with libc equivalent
3. Move sock_ctrl_msg.rs to the unix folder
4. Fix send_more_recv_less1/2() on macos (I am not sure whether this commit is correct)

### Requirements

requirements:

- [ x ] All commits in this PR have Signed-Off-By trailers (with
- [ x ] All added/changed functionality has a corresponding unit/integration
  test.
- [ x ] Any newly added `unsafe` code is properly documented.


stefano-garzarella:
@uran0sH please check the CI

uran0sH:
> @uran0sH please check the CI

Fixed

src/unix/sock_ctrl_msg.rs
```diff
@@ -253,8 +218,8 @@ unsafe fn raw_recvmsg(
                 // alignment. If these fds can not be stored in `in_fds` buffer, then all the control
                 // data must be dropped to insufficient buffer space for returning them to outer
                 // scope. This might be a sign of incorrect protocol communication.
-                for fd_offset in 0..fds_count {
-                    let raw_fds_ptr = CMSG_DATA(cmsg_ptr);
+                for fd_offset in 0..fds_to_be_copied_count {
+                    let raw_fds_ptr: *mut RawFd = CMSG_DATA(cmsg_ptr).cast();
```
germag, line 222: As discussed, this change should be in a single commit, explaining what is happening, so it will be more clear that we need help with this

stefano-garzarella: @uran0sH okay, but the point here is why:

```
When running send_more_recv_less2 or send_more_recv_less1 on macos,
cmsg.cmsg_len is 28 and CMSG_LEN(0) is 12, which cause fds_count is 4.
```

so why this is happening?

src/linux/sock_ctrl_msg.rs
```diff
@@ -535,9 +535,9 @@ mod tests {
     fn send_recv_only_fd() {
         let (s1, s2) = UnixDatagram::pair().expect("failed to create socket pair");
 
-        let evt = EventFd::new(0).expect("failed to create eventfd");
+        let mut evt = pipe().expect("failed to create pipefd");
```
stefano-garzarella, line 515: I'd suggest to use `(evt_consumer, evt_notifier)` here.

stefano-garzarella: `failed to create pipe` should be better

src/linux/sock_ctrl_msg.rs
```diff
@@ -550,21 +550,25 @@ mod tests {
...
+            .expect("Failed to read from PipeReader");
+        assert_eq!(u64::from_ne_bytes(buf), 1203);
     }
 
     #[test]
     fn send_recv_with_fd() {
         let (s1, s2) = UnixDatagram::pair().expect("failed to create socket pair");
 
-        let evt = EventFd::new(0).expect("failed to create eventfd");
+        let mut evt = pipe().expect("failed to create pipefd");
```
stefano-garzarella, line 546: Ditto.

src/unix/sock_ctrl_msg.rs
```diff
@@ -27,7 +27,13 @@ fn new_msghdr(iovecs: &mut [iovec]) -> msghdr {
         msg_name: null_mut(),
         msg_namelen: 0,
         msg_iov: iovecs.as_mut_ptr(),
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         msg_iovlen: iovecs.len(),
+        #[cfg(target_os = "macos")]
```
stefano-garzarella, line 32: Is this macos specific or can work for any unix?

src/unix/sock_ctrl_msg.rs
```diff
@@ -443,19 +437,31 @@ mod tests {
...
+            #[cfg(any(target_os = "linux", target_os = "android"))]
             assert_eq!(
                 unsafe { CMSG_SPACE(3 * size_of::<RawFd>() as c_uint) as usize },
                 size_of::<cmsghdr>() + size_of::<c_long>() * 2
             );
+            #[cfg(target_os = "macos")]
+            assert_eq!(
+                unsafe { CMSG_SPACE(3 * size_of::<RawFd>() as c_uint) as usize },
+                size_of::<cmsghdr>() + size_of::<c_uint>() * 3
+            );
```
stefano-garzarella: Can you add a comment to explain what is the difference and why?

uran0sH:
> Instead of `macos` can we refer to `unix` in the changelog/pr title/comments/etc.

`unix` is better, but I don't test it on other unix's systems

src/unix/sock_ctrl_msg.rs
uran0sH: > @uran0sH okay, but the point here is why:
> 
> ```
> When running send_more_recv_less2 or send_more_recv_less1 on macos,
> cmsg.cmsg_len is 28 and CMSG_LEN(0) is 12, which cause fds_count is 4.
> ```
> 
> so why this is happening?

I think it's the same as https://github.com/rust-vmm/vmm-sys-util/pull/245#discussion_r2209055308. The value of align causes it

stefano-garzarella:
> > Instead of `macos` can we refer to `unix` in the changelog/pr title/comments/etc.
> 
> `unix` is better, but I don't test it on other unix's systems

We can do later (e.g. in a simple freebsd vm), but yeah, please use `unix` for now

src/unix/sock_ctrl_msg.rs
stefano-garzarella: okay, so please add the reason in the commit message and maybe also in a comment in the code.

src/unix/sock_ctrl_msg.rs
stefano-garzarella: @uran0sH okay, but for the future, the "why" is the important stuff to mention in commit description, PR, etc.
We don't want each people reviewing or looking later at the commit/code, going into the details to understand the reason.

stefano-garzarella:
> Because I am not sure about this commit [7d8683c](https://github.com/rust-vmm/vmm-sys-util/commit/7d8683c35f5f26d6eb4fb5849d49db0942252be2), I think we should discuss here. Below it's what i found. linux:
> 
> ```
> CMSG_SPACE = (CMSG_ALIGN(length as usize) + CMSG_ALIGN(mem::size_of::<cmsghdr>())) as c_uint
> CMSG_ALIGN = (len + mem::size_of::<usize>() - 1) & !(mem::size_of::<usize>() - 1)
> ```
> 
> mac
> 
> ```
> CMSG_SPACE = (__DARWIN_ALIGN32(mem::size_of::<cmsghdr>()) + __DARWIN_ALIGN32(length as usize)) as c_uint
> const fn __DARWIN_ALIGN32(p: usize) -> usize {
>     const __DARWIN_ALIGNBYTES32: usize = mem::size_of::<u32>() - 1;
>     (p + __DARWIN_ALIGNBYTES32) & !__DARWIN_ALIGNBYTES32
> }
> ```
> 
> The difference is the value of align: linux uses the size of usize, and macos uses the size of u32. This is also for `CMSG_LEN`.

What are your concerns about that patch?
(maybe it's better to open the discussion, exactly on the line where the change is)

src/unix/sock_ctrl_msg.rs
stefano-garzarella, line 1: @uran0sH the [commit title](https://github.com/rust-vmm/vmm-sys-util/pull/245/commits/7d8683c35f5f26d6eb4fb5849d49db0942252be2) is not clear, here we are fixing the code, not the test.
So I think we should explain better what was the issue with previous code.

stefano-garzarella:
```
When running send_more_recv_less2 or send_more_recv_less1 on macos,
cmsg.cmsg_len is 28 and CMSG_LEN(0) is 12, which cause fds_count is 4.
```

Okay, what is the value of `fds_to_be_copied_count` and why it's fine to use it?

src/unix/sock_ctrl_msg.rs
```diff
@@ -253,8 +218,8 @@ unsafe fn raw_recvmsg(
                 // alignment. If these fds can not be stored in `in_fds` buffer, then all the control
                 // data must be dropped to insufficient buffer space for returning them to outer
                 // scope. This might be a sign of incorrect protocol communication.
-                for fd_offset in 0..fds_count {
-                    let raw_fds_ptr = CMSG_DATA(cmsg_ptr);
+                for fd_offset in 0..fds_to_be_copied_count {
```
roypat, line 221: This looks wrong to me. In this branch, we're dealing with the case of "remote send us more file descriptors than we fit into `in_fds`, thus close _all_ file descriptors that we were sent (including those of previous messages), to avoid leaking them, and return error. Changing this to `fds_to_be_copied` will only close the FDs of this message that would have fit into `in_fds`, but not the ones that don't fit.

src/unix/sock_ctrl_msg.rs
roypat: > When running send_more_recv_less2 or send_more_recv_less1 on macos,
> cmsg.cmsg_len is 28 and CMSG_LEN(0) is 12, which cause fds_count is 4.

What is the problem with this? :thinking: The test is sending 4 fds, so I would expect fds_count to be 4.

uran0sH:
> @uran0sH could you clarify what exactly test failure you were seeing with that gets fixed by [7d8683c](https://github.com/rust-vmm/vmm-sys-util/commit/7d8683c35f5f26d6eb4fb5849d49db0942252be2)? I had @ShadowCurse run the tests without that commit on his mac, and they still passed 🤔

Sorry for replying so late. It doesn't fail every time. There is a probability that it will fail the test.
How to reproduce：
```
cargo test --package vmm-sys-util --lib -- unix::sock_ctrl_msg::tests --show-output --nocapture
```
Run this command multiple times. When it fails, it reports:
```
cmsg len: 28, CMSG_LEN(0): 12
close fd: 13
close fd: 14
close fd: 0
close fd: 0


test unix::sock_ctrl_msg::tests::send_more_recv_less1 ... cmsg len: 28, CMSG_LEN: 12
close fd: 12
ok
close fd: 0
close fd: 0
close fd: 0


fatal runtime error: IO Safety violation: owned file descriptor already closed
```
(I print which fd will be closed)
Run on Linux:
```
cmsg len: 24, CMSG_LEN(0): 16
close fd: 13
close fd: 14

cmsg len: 24, CMSG_LEN(0): 16
close fd: 11
close fd: 12
```
It's so weired. Running send_more_recv_less2 or send_more_recv_less1 alone can always pass the test

src/unix/sock_ctrl_msg.rs
```diff
@@ -213,7 +207,7 @@ unsafe fn raw_recvmsg(
         let cmsg = (cmsg_ptr as *mut cmsghdr).read_unaligned();
         if cmsg.cmsg_level == SOL_SOCKET && cmsg.cmsg_type == SCM_RIGHTS {
             let fds_count: usize =
-                ((cmsg.cmsg_len - CMSG_LEN(0) as usize) as usize) / size_of::<RawFd>();
+                ((cmsg.cmsg_len as usize - CMSG_LEN(0) as usize) as usize) / size_of::<RawFd>();
```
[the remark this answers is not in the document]
uran0sH: agree

src/unix/sock_ctrl_msg.rs
```diff
@@ -635,18 +620,18 @@ mod tests {
     fn send_more_recv_less2() {
```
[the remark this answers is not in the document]
uran0sH: It looks the same. This is the original test case, so I am not sure why to write this test case

src/unix/sock_ctrl_msg.rs
```diff
@@ -119,24 +125,12 @@ fn raw_sendmsg<D: IntoIovec>(fd: RawFd, out_data: &[D], out_fds: &[RawFd]) -> Re
...
-                CMSG_LEN(
-                    size_of_val(out_fds)
-                        .try_into()
-                        .expect("size_of_val(out_fds) exceeds u32 range"),
-                ) as usize
-            },
-            #[cfg(target_env = "musl")]
+            #[cfg(all(any(target_os = "linux", target_os = "android"), not(target_env = "musl")))]
+            // SAFETY: We ensure that the input value does not exceed the range of c_uint.
+            cmsg_len: unsafe { CMSG_LEN(size_of_val(out_fds).try_into().expect("size_of_value(out_fds) exceeds u32 range")) as usize },
```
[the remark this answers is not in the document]
uran0sH: agree

src/unix/sock_ctrl_msg.rs
```diff
@@ -602,18 +587,18 @@ mod tests {
...
         let (s1, s2) = UnixDatagram::pair().expect("failed to create socket pair");
 
-        let evt1 = EventFd::new(0).expect("failed to create eventfd");
-        let evt2 = EventFd::new(0).expect("failed to create eventfd");
-        let evt3 = EventFd::new(0).expect("failed to create eventfd");
-        let evt4 = EventFd::new(0).expect("failed to create eventfd");
+        let evt1 = pipe().expect("failed to create pipefd");
+        let evt2 = pipe().expect("failed to create pipefd");
+        let evt3 = pipe().expect("failed to create pipefd");
+        let evt4 = pipe().expect("failed to create pipefd");
```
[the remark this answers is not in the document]
uran0sH: I'm not sure. Is there any problem with using Pipe here?

src/unix/sock_ctrl_msg.rs
uran0sH: > Admittedly, if we're using the libc crate's version of these, this test can probably go away altogether

It has been removed

src/unix/sock_ctrl_msg.rs
uran0sH: not sure, but change to unix now.

src/unix/sock_ctrl_msg.rs
```diff
@@ -85,34 +51,26 @@ fn new_msghdr(iovecs: &mut [iovec]) -> msghdr {
...
-    {
-        null_mut()
-    } else {
-        next_cmsg
-    }
-}
-
-const CMSG_BUFFER_INLINE_CAPACITY: usize = CMSG_SPACE!(size_of::<RawFd>() * 32);
+// SAFETY: CMSG_SPACE is a pure calculation. The input value will not exceed the range of c_uint
+const CMSG_BUFFER_INLINE_CAPACITY: usize =
```
stefano-garzarella, line 72: On my system I have this error:
```
$ cargo test
warning: `/home/stefano/repos/vmm-sys-util/.cargo/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
   Compiling vmm-sys-util v0.14.0 (/home/stefano/repos/vmm-sys-util)
error[E0015]: cannot call non-const function `CMSG_SPACE` in constants
  --> src/unix/sock_ctrl_msg.rs:73:14
   |
73 |     unsafe { CMSG_SPACE(size_of::<RawFd>() as u32 * 32) as usize };
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: calls in constants are limited to constant functions, tuple structs and tuple variants

For more information about this error, try `rustc --explain E0015`.
error: could not compile `vmm-sys-util` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `vmm-sys-util` (lib test) due to 1 previous error
```

stefano-garzarella: mmm, after a `cargo update` it worked, so not sure if we should increase the libc version, let me check

stefano-garzarella: okay, it fails with libc 0.2.126 (and previous), so I'd suggest this change:

```
diff --git a/Cargo.toml b/Cargo.toml
index 9d5c6d0..f267d85 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 with-serde = ["serde", "serde_derive"]
 
 [dependencies]
-libc = "0.2.39"
+libc = "0.2.127"
 serde = { version = "1.0.27", optional = true }
 serde_derive = { version = "1.0.27", optional = true }
```

src/unix/sock_ctrl_msg.rs
```diff
@@ -635,18 +586,18 @@ mod tests {
...
         let (s1, s2) = UnixDatagram::pair().expect("failed to create socket pair");
 
-        let evt1 = EventFd::new(0).expect("failed to create eventfd");
-        let evt2 = EventFd::new(0).expect("failed to create eventfd");
-        let evt3 = EventFd::new(0).expect("failed to create eventfd");
-        let evt4 = EventFd::new(0).expect("failed to create eventfd");
+        let (_, evt_notifier1) = pipe().expect("failed to create pipe");
+        let (_, evt_notifier2) = pipe().expect("failed to create pipefd");
+        let (_, evt_notifier3) = pipe().expect("failed to create pipefd");
+        let (_, evt_notifier4) = pipe().expect("failed to create pipefd");
```
stefano-garzarella, line 592: Ditto, please check it next time, I thought was obvious to fix in all places, no?

stefano-garzarella:
> cmsg len: 28, CMSG_LEN(0): 12
> close fd: 13
> close fd: 14
> close fd: 0
> close fd: 0
> 
> 
> test unix::sock_ctrl_msg::tests::send_more_recv_less1 ... cmsg len: 28, CMSG_LEN: 12
> close fd: 12
> ok
> close fd: 0
> close fd: 0
> close fd: 0

I think the strange thing here is have 0 as fd to close (multiple times), which is really strange.
@uran0sH When the test works, do you also have 0 as fd to close?

src/unix/sock_ctrl_msg.rs
stefano-garzarella: @roypat @uran0sH I build on macOS and replicated the issue, so the problem here is that also if the receiveir allocates a buffer of 2 fds, fds_count is set to 4 (while it should be 2), so the for loop find 2 valid fds and 2 fds set to 0, but probably because it's an overflow. MSG_CTRUNC is also set, so the fds_count calculation is wrong. Investigating....

stefano-garzarella: IMO macOS is doing something wrong with `cmsg_len`.
Our formula to calculate `fds_count` seems correct (QEMU does the same).
So, if I send 5 fds, but allocate space in the receiver for 2 fds I have this:
- Linux: set MSG_CTRUNC, set cmsg_len to have `fds_count = 2`
- macOS: set MSG_CTRUNC, set cmsg_len to have `fds_count = 5`

So it seems macOS leave it as the sender set it, which I'm not sure it's posix or not...

roypat:
Mhh, the IO violation then probably comes not from this function, but rather from the Rust "runtime" shutting down and dropping its `Stdin` object, which causes fd 0 to get closed. Which does bring up the interesting point that really, this function closing fds actually seems all sorts of wrong to me. At the very least, it should absolutely never try to close fds 0,1 and 2, because the Rust runtime definitely manages them, and closing them via libc::close() will always result in a io safety violation. But then, really, there's no way for this function to know whether some FD it received is actually managed by a `File` object elsewhere. This function should probably just return`Vec<RawFd>`, to avoid that entire problem (well, technically its unsafe so we could make the safety invariant "socket must not have more than in_fds.len() fds available", but that's not satisfiable for any caller). It still doesn't really solve the issue of macos receiving fd 0 a bunch of times though :/

src/unix/sock_ctrl_msg.rs
stefano-garzarella:
```diff
diff --git a/src/unix/sock_ctrl_msg.rs b/src/unix/sock_ctrl_msg.rs
index af1cebf..dd1352f 100644
--- a/src/unix/sock_ctrl_msg.rs
+++ b/src/unix/sock_ctrl_msg.rs
@@ -224,8 +224,8 @@ unsafe fn raw_recvmsg(
         // read.
         let cmsg = (cmsg_ptr as *mut cmsghdr).read_unaligned();
         if cmsg.cmsg_level == SOL_SOCKET && cmsg.cmsg_type == SCM_RIGHTS {
-            let fds_count: usize =
-                (cmsg.cmsg_len as usize - CMSG_LEN(0) as usize) / size_of::<RawFd>();
+            let cmsg_len: usize = std::cmp::min(cmsg.cmsg_len as usize, cmsg_capacity);
+            let fds_count: usize = (cmsg_len - CMSG_LEN(0) as usize) / size_of::<c_int>();
             // The sender can transmit more data than we can buffer. If a message is too long to
             // fit in the supplied buffer, excess bytes may be discarded depending on the type of
             // socket the message is received from.
@@ -236,7 +236,7 @@ unsafe fn raw_recvmsg(
                 // alignment. If these fds can not be stored in `in_fds` buffer, then all the control
                 // data must be dropped to insufficient buffer space for returning them to outer
                 // scope. This might be a sign of incorrect protocol communication.
-                for fd_offset in 0..fds_to_be_copied_count {
+                for fd_offset in 0..fds_count {
                     let raw_fds_ptr: *mut RawFd = CMSG_DATA(cmsg_ptr).cast();
                     // The cmsg_ptr is valid here because is checked at the beginning of the
                     // loop and it is assured to have `fds_count` fds available.
```

This fixes the issue for me on macOS, essentially limiting the cmsg_len with the cmsg_buffer we have allocated and passed to  `recvmsg`

stefano-garzarella:
@uran0sH please revert https://github.com/rust-vmm/vmm-sys-util/pull/245/commits/6a4b2131b86c3fc41a0ffa5a3fe39c73d347e80e and apply the following patch:

```patch
From f0bccfe6479937f18d231be613794d01c4f15d6f Mon Sep 17 00:00:00 2001
From: Stefano Garzarella <sgarzare@redhat.com>
Date: Fri, 18 Jul 2025 11:36:24 +0200
Subject: [PATCH] sock_ctrl_msg: fix raw_recvmsg() on macOS

From: Stefano Garzarella <sgarzare@redhat.com>

Limit cmsg_len to the size of cmsg_buffer, since it seems that in macOS,
when `MSG_CTRUNC` is set, `cmsg.cmsg_len` is not updated with the
current size, but left with the size of the sender (this is not
happening on Linux).

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>
---
 src/unix/sock_ctrl_msg.rs | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/src/unix/sock_ctrl_msg.rs b/src/unix/sock_ctrl_msg.rs
index af1cebf..f2021f3 100644
--- a/src/unix/sock_ctrl_msg.rs
+++ b/src/unix/sock_ctrl_msg.rs
@@ -224,8 +224,8 @@ unsafe fn raw_recvmsg(
         // read.
         let cmsg = (cmsg_ptr as *mut cmsghdr).read_unaligned();
         if cmsg.cmsg_level == SOL_SOCKET && cmsg.cmsg_type == SCM_RIGHTS {
-            let fds_count: usize =
-                (cmsg.cmsg_len as usize - CMSG_LEN(0) as usize) / size_of::<RawFd>();
+            let cmsg_len: usize = std::cmp::min(cmsg.cmsg_len as usize, cmsg_capacity);
+            let fds_count: usize = (cmsg_len - CMSG_LEN(0) as usize) / size_of::<c_int>();
             // The sender can transmit more data than we can buffer. If a message is too long to
             // fit in the supplied buffer, excess bytes may be discarded depending on the type of
             // socket the message is received from.
-- 
2.50.1


```

src/unix/sock_ctrl_msg.rs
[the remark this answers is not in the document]
stefano-garzarella: IIUC this is fixed in #246, so I'm resolving the conversation.

src/unix/sock_ctrl_msg.rs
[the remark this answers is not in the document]
stefano-garzarella: @roypat maybe we can fix this in #246 too (after the rebase)

stefano-garzarella:
@uran0sH okay, about this PR, the next steps are:
- [ ] please revert https://github.com/rust-vmm/vmm-sys-util/commit/6a4b2131b86c3fc41a0ffa5a3fe39c73d347e80e and apply the patch in https://github.com/rust-vmm/vmm-sys-util/pull/245#issuecomment-3088822334
- [ ] fix `expect` messages https://github.com/rust-vmm/vmm-sys-util/pull/245#discussion_r2215294227 and https://github.com/rust-vmm/vmm-sys-util/pull/245#discussion_r2215295660
- [ ] fix libc version: https://github.com/rust-vmm/vmm-sys-util/pull/245#discussion_r2215291128

If we fix those, I think we are ready to merge, so @roypat can rebase the #246 on top if this

uran0sH:
> @uran0sH great thanks! The final result LGTM, but please move the commit where to update the libc version before the commit where you use CMSG_ macros, otherwise, if in the future we will bisect this crate, we can have a build issue.

done

stefano-garzarella (approved):

roypat (approved):

